PR #6, opened 2020-06-04
Add binary sensor and remove parts from normal sensor

iMicknl:


iMicknl:
@vlebourl it won't work yet, since there are no mappings for `binary_sensors` in `TAHOMA_TYPES`. It is something we need to crowdsource. :-)

PR #13, opened 2020-06-04
Refactor cover platform

iMicknl:
Removed all hardcoded device strings, basing the supported features on the available states and commands. The main thing I am worried about is the `HorizontalAwningIOComponent` where it looks like it should be reversed, but I am not sure.

@vlebourl could you give it a try?

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,21 +30,21 @@ If your device is not supported, it will show the following message in the loggi
 
 | Somfy uiClass     | Home Assistant platform |
 | ----------------- | ----------------------- |
+| Awning            | cover                   |
 | ExteriorScreen    | cover                   |
+| GarageDoor        | cover                   |
 | Pergola           | cover                   |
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -10,11 +10,11 @@ This component is an updated version of the [original Tahoma integration](https:
 
 ### Manual
 
-Copy the `custom_components/tahoma` to your `custom_components` folder. Reboot home assistant and install the Tahoma integration via the integrations config flow.
+Copy the `custom_components/tahoma` to your `custom_components` folder. Reboot Home Assistant and install the Tahoma integration via the integrations page.
```
vlebourl, line 13: I would say "configure the Tahoma integration" instead of "install"

iMicknl: Can you give it some love in another PR? I guess there are more parts that benefit from more clear text, especially around how to install beta's and the risk.

PR #23, opened 2020-06-05
Api requests retry

vlebourl:
Added a retry logic to prevent being locked out of the api by failing gracefully after 3 attempts. 
Relates to #4.


vlebourl:
I'm not very good at managing branches... I've my climate branch in this one, that was not what I wanted to do... If you know how to clean this up, please do not hesitate!

vlebourl:
This is quite "low level" as per the component, I think this should be in a beta release of itself and tested in real life condition before merged into master.

PR #39, opened 2020-06-07
Generalize apply_action to support multiple actions, with or without parameters.

vlebourl:
accepted syntaxes are:
```
self.apply_actions("my_single_command")
self.apply_actions(["first_command_no_param", 
                               ["second_with_param", "param1", "param2"],
                               "third_command_no_param",
                             ])
```

### DIFF
--- a/custom_components/tahoma/tahoma_device.py
+++ b/custom_components/tahoma/tahoma_device.py
@@ -1,3 +1,5 @@
+from typing import List
+
 from homeassistant.helpers.entity import Entity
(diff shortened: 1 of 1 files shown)


vlebourl:
Maybe the renaming to `apply_actions` isn't necessary. Although it makes it clearer, it is a beaking change...

iMicknl:
@vlebourl is this one still needed? If yes; could we add this in a non breaking way?

vlebourl:
let's close it but keep the branch for now.

PR #45, opened 2020-06-09
Add french translation

vlebourl:


custom_components/tahoma/translations/en.json
```diff
@@ -1,22 +1,22 @@
...
-        },
-        "step": {
-            "user": {
-                "data": {
-                    "username": "Username",
-                    "password": "Password"
-                },
-                "description": "Enter the credentials you use for the Tahoma app or the tahomalink.com platform."   
-            }
+  "title": "Tahoma",
```
iMicknl, line 2: Could you remove the `title` from all files? I understood this is not mandatory and will be filled by the integration name (Somfy TaHoma). Could you try this?

vlebourl: title isn't necessary indeed. Deduced title is "Somfy Tahoma" which is good.
It seems we can't use common keys in French (and I suspect other non english languages?)
image

iMicknl: I am not sure how the common keys part works to be honest... You can translate them for now.

vlebourl: did it [here](https://github.com/iMicknl/ha-tahoma/pull/45/commits/942085efcd78a47f1e2815888261953fbaad8b56)

iMicknl:
Merci!

vlebourl:
graag gedaan

PR #46, opened 2020-06-09
renamed tahoma or Tahoma to TaHoma.

vlebourl:


custom_components/tahoma/manifest.json
```diff
@@ -1,8 +1,8 @@
 {
-  "domain": "tahoma",
+  "domain": "TaHoma",
```
iMicknl: Don't change the domain and documentation URL, since this should be all lowercase.

vlebourl: yep... I was to quick on the replace all. fixed already.

PR #48, opened 2020-06-09
Fix problem fix PositionableHorizontalAwningUno

rabesocke:
- add missing device classes to TAHOMA_HORIZONTAL_DEVICES
- add PositionableHorizontalAwningUno to TAHOMA_HORIZONTAL_DEVICES
- fix bug with reverse position state for TAHOMA_HORIZONTAL_DEVICES and unknown position for devices which only knows target states


custom_components/tahoma/const.py
```diff
@@ -56,8 +56,9 @@
 }
 
 TAHOMA_HORIZONTAL_DEVICES = {
-    "PositionableHorizontalAwning",
-    "UpDownHorizontalAwning"
+    "PositionableHorizontalAwning": DEVICE_CLASS_AWNING,
+    "PositionableHorizontalAwningUno": DEVICE_CLASS_AWNING,
+    "UpDownHorizontalAwning": DEVICE_CLASS_AWNING
```
iMicknl, line 61: The DEVICE_CLASS_AWNING is not required, it is already set in the part above by the `Awning` device type. This dict is just for enabling specific Horizontal features, because those can not be enabled based on the widget or uiclass unfortunately.

custom_components/tahoma/cover.py
```diff
@@ -121,9 +121,13 @@ def update(self):
             )
             
         if CORE_TARGET_CLOSURE_STATE in self.tahoma_device.active_states:
-            self._position = 100 - self.tahoma_device.active_states.get(
+            self._position = self.tahoma_device.active_states.get(
```
iMicknl, line 124: Why did you remove `100 -` here, but you added the `PositionableHorizontalAwningUno` to the `TAHOMA_HORIZONTAL_DEVICES`, which adds the `100-` below? You could keep the 100 - here probably, and just not add this device type to the `TAHOMA_HORIZONTAL_DEVICES`

iMicknl:
Thanks for your contribution @rabesocke! Much appreciated :)

custom_components/tahoma/const.py
rabesocke: I thought the device class is missing because it didn't work at the beginning. But probably there war only the widget type `PositionableHorizontalAwningUno` missing.

custom_components/tahoma/cover.py
rabesocke: The `PositionableHorizontalAwningUno` has to be of type `TAHOMA_HORIZONTAL_DEVICES`. If not, the awning moves in the opposite direction.

My problem at the moment is, that the HA displays the awning at both end positions (completely closed and completely open) as `closed` with `current_position = 0` when I change the code in the following way. But I don't know why - it looks as expected.

Is there a way to see the `core:TargetClosureState` without logging?

```
        if CORE_TARGET_CLOSURE_STATE in self.tahoma_device.active_states:
            self._position = 100 - self.tahoma_device.active_states.get(
                CORE_TARGET_CLOSURE_STATE
            )
            # Fix for unknown position after moving with remote control and closure states over 100
            if self._position < 0:
                self._position = self._position + 50

        if getattr(self, "_position", False):
            # HorizontalAwning devices need a reversed position that can not be obtained via the API
            if self.tahoma_device.widget in TAHOMA_HORIZONTAL_DEVICES:
                self._position = 100 - self._position

            # TODO Check if this offset is really necessary
            if self._position <= 5:
                self._position = 0
            if self._position >= 95:
                self._position = 100
```

iMicknl: If you look at ```developer-tools/state```, you can have a look at the `current_position`. However, this is not the original value since there are some modifications in the snippet above. Best way to debug is to add logging indeed.                     

One thing that is not 100% clear for me, is why you need the part below. It was something that had to be refactored anyway, because it was a bit 'ugly'. For horizontal awning devices that needs a reversed position, the code below is subtracting the position from 100, to get back to the original value. (because in the part above it has already be subtracted at the CORE_TARGET_CLOSURE_STATE). That's why I didn't understand why you had to remove the `100 -` part. 


Sorry for the mess. Is it a bit more clear now? :D

```python
     if getattr(self, "_position", False):
            # HorizontalAwning devices need a reversed position that can not be obtained via the API
            if self.tahoma_device.widget in TAHOMA_HORIZONTAL_DEVICES:
                self._position = 100 - self._position
```

custom_components/tahoma/cover.py
```diff
@@ -121,9 +121,13 @@ def update(self):
...
                 CORE_TARGET_CLOSURE_STATE
             )
+            # Fix for unknown position after moving with remote control
+            # and closure states over 100
+            if self._position > 0:
```
iMicknl, line 129: I am not sure when this could happen. The value that TaHoma returns is always between 0 and 100 I believe? Could we fix the other code to make sure don't end below 0 or above 100?

custom_components/tahoma/cover.py
rabesocke: > If you look at developer-tools/state, you can have a look at the current_position. However, this is not the original value since there are some modifications in the snippet above. Best way to debug is to add logging indeed.

Ok, than I have to do this in that way. I know the `current_position` entry in the developer section, but that's not really helpful (as you said)

> One thing that is not 100% clear for me, is why you need the part below. It was something that had to be refactored anyway, because it was a bit 'ugly'. For horizontal awning devices that needs a reversed position, the code below is subtracting the position from 100, to get back to the original value. (because in the part above it has already be subtracted at the CORE_TARGET_CLOSURE_STATE). That's why I didn't understand why you had to remove the 100 - part.

At the moment I don't know why it was necessary yesterday. I changed the code locally to the one I posted above. The problem here is (and that's a big problem), that HA set the position to `closed` for both end positions (closed and open). But I don't know why.

Maybe some more information I found out the last weeks:
- The `PositionableHorizontalAwningUno` is an extra plug in order to make normal motors to wireless motors. But this plug don't know anything about the current position. That's why these types only know something about a "target state".

- In my installation the `CORE_TARGET_CLOSURE_STATE` could be `105`. I think, that is the value when the extra plug doesn't know the state and forgot the target closure state.
So we have the problem with the initial implementation that the position is set to `-5` first and then to `105` which is then set to `100`.

I pushed the current code that runs on my installation locally at the moment (with the problem described above).

custom_components/tahoma/cover.py
rabesocke: I comment this in the other conversation and push the current solution.

custom_components/tahoma/cover.py
iMicknl: Oh the Uno part makes sense. I thought that all `io:` devices would know the state, but now the naming of TargetClosureState makes sense indeed. 

How does the online portal show the state? Do they return unknown if it is above 100? 

`is_closed` is the function you could also have a look at, they currently decide if the cover is closed based on the position in your case.

iMicknl:
@rabesocke shall I already merge those changes, so testers can benefit from your fixes. Is the current version in this branch functional at the moment?
We can have a look at cleaner code in the future.

rabesocke:
> @rabesocke shall I already merge those changes, so testers can benefit from your fixes. Is the current version in this branch functional at the moment?
> We can have a look at cleaner code in the future.

In general the current code is functional with one strange behavior (as I wrote before): the `current_state` is set to 0 when the awning is completely open. I try to find out the reason in the next days, so maybe we should wait for that.

But if you want, you can merge the current state.

iMicknl:
@rabesocke no hurry, in that case I will leave it open until you request to merge it.

vlebourl:
Hi, any update on that PR and the related issue?

rabesocke:
> Hi, any update on that PR and the related issue?

I try to find some time this weekend to find out the reason for the current state problem and to find a fix for that.

iMicknl:
@rabesocke would you have some time in the coming weeks to investigate this behaviour? We are working towards a PR for HA core, thus would be great to have this fix in.

Would be great to understand why the value can be below zero, is this because we reverse the state for Horizontal? Or is this due to incorrect value from Tahoma.

rabesocke:
The problem at the moment is that I don't have the Uno-plug anymore. The plug was only a workaround for a wrong motor in our new awning. I didn't think the engine would be replaced. Now I have the awning motor without "Uno" and can't test the fixes anymore :-(

>Would be great to understand why the value can be below zero, is this because we reverse the state for Horizontal? Or is this due to incorrect value from Tahoma.

A description of the result of my investigation is here: https://github.com/iMicknl/ha-tahoma/pull/48#discussion_r438009821

The online portal doesn't show any position for the Uno components because they don't have a position, they only have a target position. But this target position can change after a while and then it has the "undefined" value 105 (that was the situation for me).

iMicknl:
@rabesocke happy for you that you have the better motor now 👍. I will have a look at your research, it looks detailed enough for me to work on a fix, thanks!

iMicknl:
I tried fixing this in #172. Hopefully this resolves it.

PR #51, opened 2020-06-11
Added device_class to sensors

vlebourl:


custom_components/tahoma/sensor.py
```diff
@@ -63,6 +63,22 @@ def unit_of_measurement(self):
 
         return None
 
+    
+    @property
+    def device_class(self) -> Optional[str]:
+        """Return the device class of this entity if any."""
+
+        if self.tahoma_device.uiclass == "TemperatureSensor":
```
iMicknl, line 71: Nice! Could you maybe add it to const.py like the others? You can find an example in the `binary_sensor.py`.

https://github.com/iMicknl/ha-tahoma/blob/master/custom_components/tahoma/const.py

iMicknl: Do you know if the unit of measurement can be skipped when you set the device_class?

custom_components/tahoma/const.py
```diff
@@ -41,6 +43,8 @@
     "Gate": "cover"
 }
 
+TAHOMA_TYPE_HEATING_SYSTEM = "HeatingSystem"
```
iMicknl, line 46: By the way, one small comment for future PR's. What is this one used for? 

I don't see it in this PR, and I think it could be added to https://github.com/iMicknl/ha-tahoma/blob/add-device-class-to-sensors/custom_components/tahoma/const.py#L24 -> mapping `climate` in the end?

vlebourl: oups... that slipped from the climate branch... I'm still quite bad at git...

iMicknl: :D

PR #53, opened 2020-06-12
Schedule devices update right after it's added

vlebourl:
Update the sensors right after it's been added instead of waiting 1 min for the standard update.

### DIFF
--- a/custom_components/tahoma/sensor.py
+++ b/custom_components/tahoma/sensor.py
@@ -45,10 +45,6 @@ class TahomaSensor(TahomaDevice, Entity):
 
         super().__init__(tahoma_device, controller)
 
-    async def async_added_to_hass(self):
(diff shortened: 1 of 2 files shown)


custom_components/tahoma/sensor.py
```diff
@@ -45,6 +45,10 @@ def __init__(self, tahoma_device, controller):
 
         super().__init__(tahoma_device, controller)
 
+    async def async_added_to_hass(self):
```
iMicknl, line 48: Is this something that should be added to the TahomaDevice class maybe? It looks like this is also an issue for the light.py and cover.py integrations.

vlebourl: Haven't tested it on other devices than sensors but it would make sense.

PR #54, opened 2020-06-12
Add climate entity for Somfy's Smart Thermostat

vlebourl:
This is a big PR, it will need a beta release and some testing!
It also adds an `option flow` to the config flow that can be use as a basis for other options.
Relates to #12.


iMicknl:
No issues on my installation without any climate device. Would be good to change the text, or remove it, when there are no climate devices detected.

Screenshot 2020-06-16 at 13 25 00

custom_components/tahoma/config_flow.py
```diff
@@ -34,13 +41,22 @@ async def validate_input(hass: core.HomeAssistant, data):
         _LOGGER.exception("Error when trying to log in to the TaHoma API")
         raise CannotConnect
 
+    return_dict = {"title": username}
+    controller.get_setup()
```
iMicknl, line 45: Could you wrap them in the async helper?

custom_components/tahoma/config_flow.py
```diff
@@ -49,6 +65,15 @@ class ConfigFlow(config_entries.ConfigFlow, domain=DOMAIN):
     VERSION = 1
     CONNECTION_CLASS = config_entries.CONN_CLASS_CLOUD_POLL
 
+    @staticmethod
+    @callback
+    def async_get_options_flow(config_entry):
+        return ThermoOptionsFlowHandler(config_entry)
```
iMicknl, line 71: Can this be options based on the devices? or not possible here yet

custom_components/tahoma/config_flow.py
```diff
@@ -34,13 +41,22 @@ async def validate_input(hass: core.HomeAssistant, data):
...
+            return_dict[TAHOMA_TYPE_HEATING_SYSTEM][device.url] = device.label
+
     # If you cannot connect:
     # throw CannotConnect
     # If the authentication is wrong:
     # InvalidAuth
 
     # Return info that you want to store in the config entry.
-    return {"title": username}
+    return return_dict
```
iMicknl, line 59: Why is this necessary? Won't this change the name of the Somfy TaHoma entry? Currently this is the username.

vlebourl: I believe `return_dict = {"title": username}` line 44 ensures the behavior is the same as before in that regard. Am I wrong?

vlebourl: I use the return from validate_input to populate the `user_input` with elements used by climate entities (if discovered) and the options flow

custom_components/tahoma/config_flow.py
vlebourl: Not that I know of...

vlebourl:
> No issues on my installation without any climate device. Would be good to change the text, or remove it, when there are no climate devices detected.
> 
> Screenshot 2020-06-16 at 13 25 00

I changed the text if no climate device is detected.

vlebourl:
Made some substantiel changes again... I'll test those in my prod set up before marking this PR as ready for review.

PR #58, opened 2020-06-15
Add new device types

iMicknl:
- Added support for AirSensor devices
- Added support for ElectricitySensor devices
- Added support for Curtain devices
- Added support for Generic devices (cover)
- Added support for SwingingShutter devices

Based on https://github.com/dubocr/homebridge-tahoma/blob/master/mapping.json

### DIFF
--- a/custom_components/tahoma/sensor.py
+++ b/custom_components/tahoma/sensor.py
@@ -12,7 +12,9 @@ from .const import (
     TAHOMA_SENSOR_DEVICE_CLASSES,
     CORE_RELATIVE_HUMIDITY_STATE,
     CORE_LUMINANCE_STATE,
-    CORE_TEMPERATURE_STATE, 
+    CORE_TEMPERATURE_STATE,
+    CORE_ELECTRIC_POWER_CONSUMPTION_STATE,
+    CORE_CO2_CONCENTRATION_STATE
 )
 from .tahoma_device import TahomaDevice
 
@@ -55,7 +57,8 @@ class TahomaSensor(TahomaDevice, Entity):
         """Return the unit of measurement of this entity, if any."""
 
         if self.tahoma_device.uiclass == "TemperatureSensor":
-            return TEMP_CELSIUS  # TODO Retrieve core:MeasuredValueType to understand if it is Celsius or Kelvin
+            # TODO Retrieve core:MeasuredValueType to understand if it is Celsius or Kelvin
+            return TEMP_CELSIUS
 
         if self.tahoma_device.uiclass == "HumiditySensor":
             return UNIT_PERCENTAGE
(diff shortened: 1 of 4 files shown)


custom_components/tahoma/sensor.py
```diff
@@ -55,23 +57,39 @@ def unit_of_measurement(self):
...
+            return TEMP_CELSIUS
 
         if self.tahoma_device.uiclass == "HumiditySensor":
             return UNIT_PERCENTAGE
 
         if self.tahoma_device.uiclass == "LightSensor":
             return "lx"
 
+        if self.tahoma_device.uiclass == "ElectricitySensor":
+            return "W"
```
vlebourl, line 70: Is it W or Wh or kWh?

vlebourl: if the former, we should use `homeassistant.const.POWER_WATT`, 
if one of the laters, we should use `homeassistant.const.ENERGY_WATT_HOUR` or `homeassistant.const.ENERGY_KILO_WATT_HOUR`

custom_components/tahoma/sensor.py
```diff
@@ -55,23 +57,39 @@ def unit_of_measurement(self):
...
+
+        if self.tahoma_device.uiclass == "AirSensor":
+            return "ppm"
```
vlebourl, line 73: we should use `homeassistant.const.CONCENTRATION_PARTS_PER_MILLION`

custom_components/tahoma/sensor.py
```diff
@@ -86,15 +104,29 @@ def update(self):
         if CORE_RELATIVE_HUMIDITY_STATE in self.tahoma_device.active_states:
             self.current_value = float(
                 "{:.2f}".format(
```
vlebourl, line 106: I had a review on a PR in core asking to use f-strings. should we use them instead of `"".format`?

vlebourl: `f"{self.tahoma_device.active_states.get(CORE_RELATIVE_HUMIDITY_STATE):.2f}"`

iMicknl: Not sure, I have seen the comment indeed, but this is how I retrieved it from the original and other components. We could change all of those .format to fstrings in another refactor PR.

vlebourl: fine by me.

vlebourl: https://developers.home-assistant.io/docs/development_guidelines/#use-new-style-string-formatting

iMicknl:
```
self.current_value = f"{self.tahoma_device.active_states.get(CORE_RELATIVE_HUMIDITY_STATE):.2f}"
```

I am not sure if this is more readable in this case, so I will leave it for now.

PR #70, opened 2020-06-17
adds async_apply_action

vlebourl:


vlebourl:
although it's not really needed, it helps keep track of the syntax, and might be useful in the future, I suggest to merge it for future ref.

Issue #73, opened 2020-06-19
Add support for IO RGB Light (io:DimmableRGBLightIOComponent)


jurgen272:
IO RGB Light.
ON/Off and dimming is working at this moment, but I can't change the colors

**Device details**

```json
{
         "creationTime":1590511973000,
         "lastUpdateTime":1590511973000,
         "label":"Light Terras RGB",
         "deviceURL":"io://xxxxxxx",
         "shortcut":false,
         "controllableName":"io:DimmableRGBLightIOComponent",
         "definition":{
            "commands":[
               {
                  "commandName":"delayedStopIdentify",
                  "nparams":1
               },
               {
                  "commandName":"getName",
                  "nparams":0
               },
               {
                  "commandName":"identify",
                  "nparams":0
               },
               {
                  "commandName":"my",
                  "nparams":0
               },
               {
                  "commandName":"off",
                  "nparams":0
               },
               {
                  "commandName":"on",
                  "nparams":0
               },
               {
                  "commandName":"onWithTimer",
                  "nparams":1
               },
               {
                  "commandName":"refreshMemorized1Position",
                  "nparams":0
               },
               {
                  "commandName":"setIntensity",
                  "nparams":1
               },
               {
                  "commandName":"setIntensityWithTimer",
                  "nparams":3
               },
               {
                  "commandName":"setMemorized1Position",
                  "nparams":1
               },
               {
                  "commandName":"setName",
                  "nparams":1
               },
               {
                  "commandName":"setOnOff",
                  "nparams":1
               },
               {
                  "commandName":"setPosition",
                  "nparams":1
               },
               {
                  "commandName":"setRGB",
                  "nparams":3
               },
               {
                  "commandName":"setRGBWithTimer",
                  "nparams":4
               },
               {
                  "commandName":"setSecuredPosition",
                  "nparams":1
               },
               {
                  "commandName":"startIdentify",
                  "nparams":0
               },
               {
                  "commandName":"stopIdentify",
                  "nparams":0
               },
               {
                  "commandName":"wink",
                  "nparams":1
               }
            ],
            "states":[
               {
                  "type":"ContinuousState",
                  "qualifiedName":"core:BlueColorIntensityState"
               },
               {
                  "type":"ContinuousState",
                  "qualifiedName":"core:GreenColorIntensityState"
               },
               {
                  "type":"ContinuousState",
                  "qualifiedName":"core:LightIntensityState"
               },
               {
                  "type":"ContinuousState",
[...]

PR #75, opened 2020-06-19
Add support for "setRGB" command for light entity.

vlebourl:
fix issue #73

### DIFF
--- a/custom_components/tahoma/light.py
+++ b/custom_components/tahoma/light.py
@@ -94,15 +94,16 @@ class TahomaLight(TahomaDevice, LightEntity):
         self._state = True
 
         if ATTR_HS_COLOR in kwargs:
-            self._hs_color = kwargs[ATTR_HS_COLOR]
             self.apply_action(
                 "setRGB",
-                *[int(float(c)) for c in color_util.color_hs_to_RGB(*self._hs_color)],
+                *[
+                    int(float(c))
+                    for c in color_util.color_hs_to_RGB(*kwargs[ATTR_HS_COLOR])
+                ],
             )
-
         if ATTR_BRIGHTNESS in kwargs:
-            self._brightness = int(float(kwargs[ATTR_BRIGHTNESS]) / 255 * 100)
-            self.apply_action("setIntensity", self._brightness)
+            brightness = int(float(kwargs[ATTR_BRIGHTNESS]) / 255 * 100)
+            self.apply_action("setIntensity", brightness)
         elif ATTR_EFFECT in kwargs:
             self._effect = kwargs[ATTR_EFFECT]
             self.apply_action("wink", 100)
--- a/custom_components/tahoma/light.py
+++ b/custom_components/tahoma/light.py
@@ -103,7 +103,7 @@ class TahomaLight(TahomaDevice, LightEntity):
         _LOGGER.warning(f"light.turn_on kwargs: {kwargs}")
 
         if ATTR_HS_COLOR in kwargs:
-            rgb = color_util.color_hs_to_RGB(kwargs[ATTR_HS_COLOR])
+            rgb = color_util.color_hs_to_RGB(*kwargs[ATTR_HS_COLOR])
             self._rgb = [int(float(c)) for c in rgb]
             _LOGGER.warning(f"self._rgb: {self._rgb}")
             self._apply_action("setRGB", *self._rgb)
--- a/custom_components/tahoma/light.py
+++ b/custom_components/tahoma/light.py
@@ -5,7 +5,9 @@ import logging
 from homeassistant.components.light import (
     ATTR_BRIGHTNESS,
     ATTR_EFFECT,
+    ATTR_RGB_COLOR,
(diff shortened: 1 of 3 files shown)


iMicknl:
Don't we need to pull this from the states as well? To show the current color.

vlebourl:
Indeed we do. Added in the last commit.

custom_components/tahoma/light.py
```diff
@@ -68,13 +85,22 @@ def supported_features(self) -> int:
...
 
     def turn_on(self, **kwargs) -> None:
         """Turn the light on."""
         self._state = True
         self._skip_update = True
 
+        _LOGGER.debug(f"light.turn_on kwargs: {kwargs}")
+
+        if ATTR_RGB_COLOR in kwargs:
+            self._rgb = [int(float(c)) for c in kwargs[ATTR_RGB_COLOR]]
```
iMicknl, line 101: Could you log self._rgb? you need 3 params in `setRGB`, I believe this could be the culprit

vlebourl: done.
but if kwargs[ATTR_RGB_COLOR] is a 3 elements list, which the code suggests, then that should work: `self.apply_action("setRGB", *self._rgb)`

custom_components/tahoma/light.py
```diff
@@ -68,29 +85,44 @@ def supported_features(self) -> int:
...
     def turn_on(self, **kwargs) -> None:
         """Turn the light on."""
         self._state = True
-        self._skip_update = True
+
+        _LOGGER.warning(f"light.turn_on kwargs: {kwargs}")
+
+        if ATTR_HS_COLOR in kwargs:
+            rgb = color_util.color_hs_to_RGB(*kwargs[ATTR_HS_COLOR])
+            self._rgb = [int(float(c)) for c in rgb]
```
iMicknl, line 107: Can we already set `self._rgb` here? What if the command fails, in that case the state is incorrect? 

Or should we implement this behavior in all integrations, where we directly update the state, but change it back if the command fails. (which happens via the update() function)

custom_components/tahoma/light.py
```diff
@@ -56,6 +66,13 @@ def is_on(self) -> bool:
         """Return true if light is on."""
         return self._state
 
+    @property
+    def hs_color(self):
+        """Return the hue and saturation color value [float, float]."""
+        if self.supported_features & SUPPORT_COLOR:
```
iMicknl, line 72: Maybe better to check if `self._rgb` is set? This function will only be called if `SUPPORT_COLOR` is in `supported_features`.

vlebourl: It seems from `kwargs` given to `turn_on` and the `hs_color` property that RGB isn't the standard way. Should we implement hue saturation instead of RGB, and only convert to RGB for the apply_action which is specific to TaHoma?

custom_components/tahoma/light.py
```diff
@@ -68,29 +85,43 @@ def supported_features(self) -> int:
...
 
+        if "setRGB" in self.tahoma_device.command_definitions:
+            supported_features |= SUPPORT_COLOR
+
         return supported_features
 
+    def _apply_action(self, cmd_name, *args):
+        """Apply an action and wait for it to complete."""
+        exec_id = self.apply_action(cmd_name, *args)
+        while exec_id in self.controller.get_current_executions():
```
iMicknl, line 96: This wait method has been added to TahomaDevice.

custom_components/tahoma/light.py
```diff
@@ -125,3 +152,10 @@ def update(self):
             self._state = True
         else:
             self._state = False
+
+        if CORE_RED_COLOR_INTENSITY_STATE in self.tahoma_device.active_states:
+            self._rgb = [
```
iMicknl, line 157: `self._rgb` -> `self._hs`

custom_components/tahoma/light.py
iMicknl: This is one still applies, but is now renamed to `self._hs_color`

custom_components/tahoma/light.py
```diff
@@ -5,13 +5,23 @@
 from homeassistant.components.light import (
     ATTR_BRIGHTNESS,
     ATTR_EFFECT,
+    ATTR_HS_COLOR,
+    ATTR_RGB_COLOR,
```
iMicknl, line 9: Unused

custom_components/tahoma/light.py
vlebourl: Agreed, `self._hs_color` should be set in the update triggered after the turn_on, now that the command awaits for its completion.

vlebourl: isn't that the same for `_brightness` and `_effect` though?

iMicknl: You are right, let's keep it for now to have the same behavior across this integration.

vlebourl: ok, let me push that and we can merge.

PR #78, opened 2020-06-21
Update cover & changelog

iMicknl:


custom_components/tahoma/cover.py
```diff
@@ -318,12 +318,12 @@ def stop_cover(self, **kwargs):
         if "stop" in self.tahoma_device.command_definitions:
             return self.apply_action("stop")
 
-        if "my" in self.tahoma_device.command_definitions:
-            return self.apply_action("my")
-
         if "stopIdentify" in self.tahoma_device.command_definitions:
             return self.apply_action("stopIdentify")
 
+        if "my" in self.tahoma_device.command_definitions:
```
vlebourl, line 324: Are we sure 'my' is lower priority?

iMicknl: Not 100% to be honest. However according to the original component, my is used less frequently. I can withdraw this 'fix' from this PR if you want.

Issue #80, opened 2020-06-22
Electricity Meter


lseufer:
Describe the bug:
Data isn't pulled from my meters

Expected behavior:
Nothing is reported. Only "unknown" 

Environment:
 - Home Assistant version: 0.111
 - **Platform**: sensor

Device:
 - Model: 'io:TotalElectricalEnergyConsumptionIOSystemSensor'
 - Type: CumulativeElectricPowerConsumptionSensor

vlebourl:
Hi, thank you for the report, can you please follow instructions provided [here](https://github.com/iMicknl/ha-tahoma/blob/7d8a780b9a3d0ac6b07e76967dbe3d8b36beb49b/.github/ISSUE_TEMPLATE/bug_report.md) to add device details, namely state definitions and command definitions?

lseufer:
Here you go!


`creationTime":1564688966000,
         "lastUpdateTime":1564688966000,
         "label":"Compteur",
         "deviceURL":"",
         "shortcut":false,
         "controllableName":"io:TotalElectricalEnergyConsumptionIOSystemSensor",
         "definition":{
            "commands":[

            ],
            "states":[
               {
                  "type":"ContinuousState",
                  "qualifiedName":"core:ElectricEnergyConsumptionState"
               },
               {
                  "type":"ContinuousState",
                  "qualifiedName":"core:RSSILevelState"
               },
               {
                  "values":[
                     "dead",
                     "lowBattery",
                     "maintenanceRequired",
                     "noDefect"
                  ],
                  "type":"DiscreteState",
                  "qualifiedName":"core:SensorDefectState"
               },
               {
                  "values":[
                     "available",
                     "unavailable"
                  ],
                  "type":"DiscreteState",
                  "qualifiedName":"core:StatusState"
               }
            ],
            "dataProperties":[

            ],
            "widgetName":"CumulativeElectricPowerConsumptionSensor",
            "uiClass":"ElectricitySensor",
            "qualifiedName":"io:TotalElectricalEnergyConsumptionIOSystemSensor",
            "type":"SENSOR"
         },
         "states":[
            {
               "name":"core:StatusState",
               "type":3,
               "value":"available"
            },
            {
               "name":"core:ElectricEnergyConsumptionState",
               "type":1,
               "value":21744148
            }
         ],
         "attributes":[
            {
               "name":"core:MeasurementCategory",
               "type":3,
               "value":"electric/overall"
            },
            {
[...]

PR #81, opened 2020-06-22
[feat] Support io:TotalElectricalEnergyConsumptionIOSystemSensor

vlebourl:
Related Github issues: #80

### DIFF
--- a/custom_components/tahoma/sensor.py
+++ b/custom_components/tahoma/sensor.py
@@ -6,6 +6,7 @@ from typing import Optional
 from homeassistant.const import (
     ATTR_BATTERY_LEVEL,
     CONCENTRATION_PARTS_PER_MILLION,
+    ENERGY_KILO_WATT_HOUR,
     POWER_WATT,
     TEMP_CELSIUS,
     UNIT_PERCENTAGE,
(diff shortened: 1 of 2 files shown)


custom_components/tahoma/sensor.py
```diff
@@ -72,9 +74,20 @@ def unit_of_measurement(self):
         if self.tahoma_device.uiclass == "LightSensor":
             return "lx"
 
-        if self.tahoma_device.uiclass == "ElectricitySensor":
+        if (
+            self.tahoma_device.uiclass == "ElectricitySensor"
```
iMicknl, line 78: Is `ElectricitySensor` necessary here? I like your approach even more, since it can be more modular.

custom_components/tahoma/sensor.py
```diff
@@ -130,6 +143,18 @@ def update(self):
...
+        if CORE_ELECTRIC_ENERGY_CONSUMPTION_STATE in self.tahoma_device.active_states:
+            self.current_value = (
+                float(
+                    "{:.2f}".format(
+                        self.tahoma_device.active_states.get(
+                            CORE_ELECTRIC_ENERGY_CONSUMPTION_STATE
+                        )
+                    )
+                )
+                / 1000
```
iMicknl, line 155: Any reason why you chose kW over W (which is the value returned by TaHoma?)

vlebourl: Energy is usually given in kWh rather than Wh, which also improves the readability.

PR #84, opened 2020-06-22
Add support for more sensors & basic Siren (switch) support

iMicknl:
### Added

- Added initial support for AirFlowSensor
- Added initial support for WaterDetectionSensor
- Added support for RainSensor
- Added support for CarButtonSensor
- Added basic support for Siren as switch
- Added support for MotionSensor
- Added support for SunSensor
- Added support for WindSensor
- Added all available states to device state attributes
- Added support for COSensor and CO2Sensor (AirSensor)

### Changed

- Change 'battery_level' attribute to an integer, estimated based on the battery state
- Changed CO2Sensor device class to co2 to show as co2 sensor in HomeKit  
- Remove tahoma_scene_oid, which is already used as unique_id for scenes


CHANGELOG.md
```diff
@@ -2,11 +2,20 @@
 
 ## [1.5](https://github.com/iMicknl/ha-tahoma/tree/1.5) (2020-07-03)
```
iMicknl, line 4: @vlebourl how should I do the current changelog approach?

vlebourl: I generate the changelog using https://github.com/github-changelog-generator/github-changelog-generator
But it needs the PR to be merged to integrate the changes in the changelog. 
I propose to keep the changelog out of the PRs in general, and after a successful merge, merge master into branch https://github.com/iMicknl/ha-tahoma/tree/update-changelog and update the changelog.

README.md
```diff
@@ -28,44 +28,36 @@ If your device is not supported, it will show the following message in the debug
...
-| TemperatureSensor | sensor                  |
-| DoorLock          | lock                    |
-| OnOff             | switch                  |
-| ContactSensor     | binary_sensor           |
-| OccupancySensor   | binary_sensor           |
-| SmokeSensor       | binary_sensor           |
-| WindowHandle      | binary_sensor           |
-| Light             | light                   |
-
-## Not supported (yet)
```
vlebourl, line 56: I see you removed this section, do we support those now? or is the list too long so we just don't specify what is not supported?

custom_components/tahoma/binary_sensor.py
```diff
@@ -58,6 +64,21 @@ def device_class(self):
             or None
         )
 
+    @property
+    def icon(self) -> Optional[str]:
+        """Return the icon to use in the frontend, if any."""
+
+        if self.device_class == DEVICE_CLASS_RAIN:
+            return "mdi:weather-rainy"
```
vlebourl, line 72: can we use `hass` icons instead of `mdi`? I think they are all the same, but load faster.
`hass:weather-rainy`

custom_components/tahoma/binary_sensor.py
```diff
@@ -66,21 +87,25 @@ def update(self):
...
+            self.current_value = states.get(CORE_SMOKE_STATE) == "detected"
+
+        if CORE_RAIN_STATE in states:
+            self.current_value = states.get(CORE_RAIN_STATE) == "detected"
+
+        if CORE_WATER_DETECTION_STATE in states:
+            self.current_value = states.get(CORE_WATER_DETECTION_STATE) == "detected"
+
+        if IO_VIBRATION_STATE in states:
+            self.current_value = states.get(IO_VIBRATION_STATE) == "detected"
```
vlebourl, line 108: Should we have constants in the beginning of the file or in `const.py` instead of hardcoded words here for the ones not in `core/const.py`?

custom_components/tahoma/sensor.py
```diff
@@ -89,9 +96,15 @@ def unit_of_measurement(self):
     def icon(self) -> Optional[str]:
         """Return the icon to use in the frontend, if any."""
 
-        if self.tahoma_device.uiclass == "AirSensor":
+        if self.device_class == DEVICE_CLASS_CO2:
             return "mdi:periodic-table-co2"
```
vlebourl, line 100: use hass icons?

custom_components/tahoma/sensor.py
```diff
@@ -111,47 +124,43 @@ def update(self):
...
-        if CORE_RELATIVE_HUMIDITY_STATE in self.tahoma_device.active_states:
+        if CORE_LUMINANCE_STATE in states:
+            self.current_value = states.get(CORE_LUMINANCE_STATE)
+
+        if CORE_RELATIVE_HUMIDITY_STATE in states:
             self.current_value = float(
-                "{:.2f}".format(
-                    self.tahoma_device.active_states.get(CORE_RELATIVE_HUMIDITY_STATE)
-                )
+                "{:.2f}".format(states.get(CORE_RELATIVE_HUMIDITY_STATE))
```
vlebourl, line 134: We should use f-strings here. I think this will be required for a merge in core.
`f'{states.get(CORE_RELATIVE_HUMIDITY_STATE):%.2f}'`

custom_components/tahoma/switch.py
```diff
@@ -31,47 +32,66 @@ class TahomaSwitch(TahomaDevice, SwitchEntity):
...
-        if self._skip_update:
-            self._skip_update = False
-            return
-
         self.controller.get_states([self.tahoma_device])
 
-        _LOGGER.debug("Update %s, state: %s", self._name, self._state)
+        if CORE_ON_OFF_STATE in self.tahoma_device.active_states:
+            self.current_value = (
+                self.tahoma_device.active_states.get(CORE_ON_OFF_STATE) == "on"
```
vlebourl, line 49: use STATE_ON

custom_components/tahoma/switch.py
```diff
@@ -31,47 +32,66 @@ class TahomaSwitch(TahomaDevice, SwitchEntity):
...
+    @property
+    def icon(self) -> Optional[str]:
+        """Return the icon to use in the frontend, if any."""
+
+        if self.device_class == DEVICE_CLASS_SIREN:
+            if self.is_on:
+                return "mdi:bell-ring"
+            else:
+                return "mdi:bell-off"
+
```
vlebourl, line 70: use `hass:` icons

custom_components/tahoma/switch.py
```diff
@@ -31,47 +32,66 @@ class TahomaSwitch(TahomaDevice, SwitchEntity):
...
 
-        self.apply_action("on")
-        self._skip_update = True
-        self._state = STATE_ON
+        if "on" in self.tahoma_device.command_definitions:
+            return self.apply_action("on")
+
+        if "ringWithSingleSimpleSequence" in self.tahoma_device.command_definitions:
+            return self.apply_action(
+                "ringWithSingleSimpleSequence", 120000, 75, 2, "memorizedVolume"
```
vlebourl, line 81: These are adhoc values taken from the doc if I recall correctly. Do we agree to keep those until we understand better what it does?

README.md
iMicknl: We missed a lot of element in Not Supported, and since we are able to retrieve all the possible elements via the API now, I thought it didn't really make sense to document them here. I added them here as a reference.

custom_components/tahoma/binary_sensor.py
iMicknl: Is there any documentation around this? I am not aware of the icons changing to hass prefix?

custom_components/tahoma/switch.py
iMicknl: `STATE_ON` is a Home Assistant const, where "on" is a state from Tahoma. What if the Home Assistant const changes, but Tahoma not? I would like to keep them separate, however maybe good to have on / off / detected etc. also in the Tahoma const?

README.md
vlebourl: Ok, let's take this part out.

custom_components/tahoma/binary_sensor.py
vlebourl: I can't find where I've read this... Let's keep mdi.

custom_components/tahoma/switch.py
vlebourl: Fine with me to keep them separate. But I think it would be better to have those as const, and try to not use hardcoded string outside of const.py or the header of integration files.

custom_components/tahoma/binary_sensor.py
iMicknl: I am not sure yet, to be honest. I like to have at least everything defined as a const, also the "detected" strings at some point. (future PR). However not sure if we should try to have the constants in the platform files and only shared constants in `consts.py`, or if we should keep every constant in a single file.

@tetienne do you have an opinion on this?

vlebourl: ok. Let's not keep this from merging for now. We can make an enhancement PR later regarding constants

custom_components/tahoma/switch.py
iMicknl: Is there a risk of us breaking devices? If not, we could maybe keep this in and listen to feedback at some point. Or we could disable the Siren as a switch in the const for now.

custom_components/tahoma/binary_sensor.py
tetienne: It really depends. If the keyword "detected" is only related to the binary_sensor platform, put the constant here, else in constants.py.

iMicknl: @tetienne @vlebourl lets do that in a future PR indeed. There are many constants in const.py that are only used for a single platform at the moment, thus this will be some work :O

custom_components/tahoma/switch.py
vlebourl: I don't think it can break anything, aside from someone's ears...

tetienne:
Not related to the code, but I highly recommand you do only squash and merge your PR. You will have a nice commit history in the master branch without all your WIP and Fix commit of your branch. 
Many projects do like this, HA included.

custom_components/tahoma/sensor.py
```diff
@@ -89,9 +102,18 @@ def unit_of_measurement(self):
...
+
+        if self.device_class == DEVICE_CLASS_CO2:
             return "mdi:periodic-table-co2"
 
+        if self.device_class == DEVICE_CLASS_WIND_SPEED:
+            return "mdi:weather-windy"
+
+        if self.device_class == DEVICE_CLASS_SUN_ENERGY:
+            return "mdi:solar-power"
+
```
tetienne, line 116:
```suggestion
 icons = {
    DEVICE_CLASS_CO: "mdi:air-filter",
    DEVICE_CLASS_CO2: "mdi:periodic-table-co2",
    DEVICE_CLASS_WIND_SPEED: "mdi:weather-windy",
    DEVICE_CLASS_SUN_ENERGY:"mdi:solar-power"
}
return icons.get(self.device_class)
```

custom_components/tahoma/switch.py
tetienne: It would be really nice to have a comment to explain where these values come from.

iMicknl:
> Not related to the code, but I highly recommand you do only squash and merge your PR. You will have a nice commit history in the master branch without all your WIP and Fix commit of your branch.
> Many projects do like this, HA included.

Makes sense. Especially since this PR doesn't include a nice commit log. 👍

custom_components/tahoma/sensor.py
tetienne: It will reduce the complexity of your method.

tetienne:
> > Not related to the code, but I highly recommand you do only squash and merge your PR. You will have a nice commit history in the master branch without all your WIP and Fix commit of your branch.
> > Many projects do like this, HA included.
> 
> Makes sense. Especially since this PR doesn't include a nice commit log. +1

You got it. You can allow only this method if you want in the settings of your repository.

custom_components/tahoma/binary_sensor.py
tetienne: Do you know if the devices related to the "detected" cases have something in common? For instance the widget?

custom_components/tahoma/sensor.py
iMicknl: @tetienne do we need fstrings for all floats here? https://github.com/home-assistant/core/pull/36258#issuecomment-637720333

custom_components/tahoma/binary_sensor.py
iMicknl: @tetienne no, unfortunately not. That's why we need a mapping for Sensors to understand if it is a `binary_sensor` or a `sensor`. However at the moment, we support almost all available sensors.

iMicknl: However, there are a few files with a lot of if statements now, that should be mapped using some helper functions. There has been some progress in cover for the commands, but it is something that could be done for states as well.

iMicknl: https://github.com/iMicknl/ha-tahoma/pull/91/files <-- we should add a helper function like this at some point, for states and commands.

custom_components/tahoma/sensor.py
tetienne: What do you have in entry and what do you want in output?

To have a float from a string you can do:
```
>>>a = "2.3"
>>>float(a)
2.3
```

custom_components/tahoma/binary_sensor.py
tetienne: Indeed this pattern occurred a lot.

tetienne: > @tetienne no, unfortunately not. That's why we need a mapping for Sensors to understand if it is a `binary_sensor` or a `sensor`. However at the moment, we support almost all available sensors.

Can you give me a device which has this CORE_WATER_DETECTION_STATE state and other which has CORE_GAS_DETECTION_STATE?

custom_components/tahoma/sensor.py
vlebourl: do we actually convert a float to string then to float?

custom_components/tahoma/binary_sensor.py
iMicknl: `rtds:RTDSGasSensor` and `rtds:RTDSWaterSensor`

iMicknl: > Indeed this pattern occurred a lot.

This is mainly due to the fact that it started small, but at the moment many devices are supported, thus asking for some refactors throughout the component.

iMicknl: Logged an issue here; https://github.com/iMicknl/ha-tahoma/issues/110

custom_components/tahoma/sensor.py
iMicknl: To be honest, I am not sure how HA would like to have the values. Do they require a float or is a floating value in a string the right approach?

In the end, what should happen in `sensor.py` is that we make round the value and make it an int or float.

vlebourl: I think a float is good enough. but why not do `round(states.get(CORE_RELATIVE_HUMIDITY_STATE),2)` instead of converting to a string for the rounding?

vlebourl: and I would do the same in sensor.py

tetienne: @vlebourl, is states.get(CORE_RELATIVE_HUMIDITY_STATE) already a float? If a string, the round method can not be applied.
```
>>> a = "2.3"
>>> round(a, 2)
  File "<stdin>", line 1, in <module>
TypeError: type str doesn't define __round__ method

Comment (vlebourl) [reply custom_components/tahoma/sensor.py:138]: I think it. otherwise we would get
```
>>> "{:.2f}".format("2.45")
  File "<input>", line 1, in <module>
ValueError: Unknown format code 'f' for object of type 'str'
```

Comment (vlebourl) [reply custom_components/tahoma/sensor.py:138]: And that's the signature for my temperature sensor:
image

Comment (vlebourl) [reply custom_components/tahoma/sensor.py:138]: Let's maybe leave this like this for now as this PR is already quite large, but resolve this in another PR once this one is merged

**Comment by vlebourl** [comment]
Shall we merge?

File: custom_components/tahoma/sensor.py (hunk 6 above)
Comment (iMicknl) [reply custom_components/tahoma/sensor.py:138]: Agree, let's work on those improvements in separate smaller PR's. This one already got way to big :D

**Comment by tetienne** [comment]
Exemple of water sensor
```
{
      "typeId": 131,
      "subsystemId": 0,
      "localPairing": false,
      "states": [
        {
          "name": "SensorDefectState",
          "type": "DiscreteState",
          "eventBased": false,
          "persistent": false
        },
        {
          "name": "WaterDetectionState",
          "type": "DiscreteState",
          "eventBased": true,
          "persistent": false,
          "prototype": {
            "valuePrototypes": [
              {
                "type": "STRING",
                "enumValues": [
                  "detected",
                  "notDetected"
                ],
                "description": "string value (detected, notDetected)"
              }
            ]
          }
        }
      ],
      "attributes": [
        {
          "type": 3,
          "defaultValue": "battery"
        }
      ],
      "controllableName": "rtds:RTDSWaterSensor",
      "uiClass": "HumiditySensor",
      "uiProfiles": [
        "WaterDetector"
      ],
      "uiWidget": "WaterDetectionSensor",
      "controllableType": "SENSOR",
      "protocolType": "RTDS"
    }
```

PR #91, opened 2020-06-29
[Refactor] Add helper functions select_command and select_state.

vlebourl:
Add two functions to select first existing command or state from a list.
Implemented in cover.py as a proof of concept.


vlebourl:
That was a PR to refactor and simplified code. Not urgent for now.

Issue #92, opened 2020-06-29
Actions are applied sequentialy


vlebourl:
**Describe the bug**
When triggering multiple actions in a short time period, actions are executed sequentially by the box instead of concurrently.

**Expected behavior**
Action should be able to execute at once on independent devices.

**Additional context**
This probably relates to the introduction of this part in apply_action:
```
        while exec_id in self.controller.get_current_executions():
            continue
```

This relates to #90 (2)

vlebourl:
This has been tested with covers on my end, but would need further testing, and might not be the best solution...
I guess rewriting the API with event based updates would also render this outdated. 
Maybe simply reverting the waiting in the apply_action is enough for now? I think it is useful that entities that take time to apply an action (like closing covers or gates, activating a particular mode on a heater) wait until the action is completed before being updated so that the UI reflects the actual state of the device, and doesn't have to wait for `SCAN_INTERVAL` before having an accurate state displayed.

Issue #90, opened 2020-06-27
Covers move step by step and can't be stopped while moving


rabesocke:
**Describe the bug**
I have two problems with version 1.4. At the moment I switched back to version 1.3.

- The covers can't be stopped while they are moving. They move all the way down and don't respond to the stop button.
- By activating a scene with several covers all covers move one by one. The first cover moves and when that one is finished the next cover moves. Until version 1.3 all covers moves all together.

**To Reproduce**
1. Move a cover and click the stop button - the cover do not respond to that click

or

1. Create a scene with several covers.
2. Activate the scene

**Expected behavior**
- The cover stops by clicking the stop button while moving
- All covers should move together (and not one by one)

**Screenshots**
No screenshot

**Environment (please complete the following information):**
 - **Home Assistant version:** 0.111.4
 - **Platform**: cover

**Device:** (if your problem is related to a specific device)
 - **Model**: PositionableRollerShutter
 - **Type**: io:RollerShutterGenericIOComponent
 - **Device details**
no further details

**Additional context**
No additional context necessary

vlebourl:
thank you for your report. we'll have a look rapidly. 
@Mick regarding the second issue, I think this is related to the addition of the wait to the apply_action. could an async function allow to keep the wait while sending several commands without waiting for the previous ones to finish?

vlebourl:
small comment, it would be preferable to open 2 separate issues. Thanks a lot :)

Issue #86, opened 2020-06-24
Not authenticated error.


vlebourl:
**Describe the bug**
For some unknown reason, we get kicked out of the api, causing the following error message:
`{"errorCode":"RESOURCE_ACCESS_DENIED","error":"Not authenticated"}`

**To Reproduce**
Don't know yet...

mariwing:
This happens a lot to me, sometimes already after a restart, sometimes after one or more day. This is the complete error message I get flooded (litterally 100 of these entries):

2020-07-01 21:05:09 ERROR (MainThread) [homeassistant.helpers.entity] Update for cover.markise_hoyre fails
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 272, in async_update_ha_state
    await self.async_device_update()
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 466, in async_device_update
    self.update  # type: ignore
  File "/usr/local/lib/python3.7/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/config/custom_components/tahoma/cover.py", line 92, in update
    self.controller.get_states([self.tahoma_device])
  File "/config/custom_components/tahoma/tahoma_api.py", line 513, in get_states
    result = self.send_request(requests.get, BASE_URL + path, headers=header)
  File "/config/custom_components/tahoma/tahoma_api.py", line 120, in send_request
    self.send_request(method, url, headers, data, timeout, retries - 1)
  File "/config/custom_components/tahoma/tahoma_api.py", line 120, in send_request
    self.send_request(method, url, headers, data, timeout, retries - 1)
  File "/config/custom_components/tahoma/tahoma_api.py", line 120, in send_request
    self.send_request(method, url, headers, data, timeout, retries - 1)
  File "/config/custom_components/tahoma/tahoma_api.py", line 117, in send_request
    + request.text
Exception: Maximum number of consecutive retries reached. Error is:
{"errorCode":"RESOURCE_ACCESS_DENIED","error":"Not authenticated"}

vlebourl:
Thanks for the report. Unfortunately I haven't found a way to reproduce it consistently to debug it... Still looking!

vlebourl:
[...]

PR #93, opened 2020-06-29
Remove the lock on apply_action

vlebourl:
#90
#92

### DIFF
--- a/custom_components/tahoma/binary_sensor.py
+++ b/custom_components/tahoma/binary_sensor.py
@@ -60,6 +60,12 @@ class TahomaBinarySensor(TahomaDevice, BinarySensorEntity):
 
     def update(self):
         """Update the state."""
+        exec_queue = self.controller.get_current_executions()
+        self._exec_queue = [e for e in self._exec_queue if e in exec_queue]
+        if self._exec_queue:
+            self.schedule_update_ha_state(True)
+            return
+
         self.controller.get_states([self.tahoma_device])
 
         if CORE_CONTACT_STATE in self.tahoma_device.active_states:
--- a/custom_components/tahoma/binary_sensor.py
+++ b/custom_components/tahoma/binary_sensor.py
@@ -60,9 +60,7 @@ class TahomaBinarySensor(TahomaDevice, BinarySensorEntity):
 
     def update(self):
         """Update the state."""
-        exec_queue = self.controller.get_current_executions()
-        self._exec_queue = [e for e in self._exec_queue if e in exec_queue]
-        if self._exec_queue:
+        if self.should_wait():
             self.schedule_update_ha_state(True)
             return
 
--- a/custom_components/tahoma/climate.py
+++ b/custom_components/tahoma/climate.py
@@ -299,6 +299,12 @@ class TahomaClimate(TahomaDevice, ClimateEntity):
 
     def update(self):
         """Update the state."""
+        exec_queue = self.controller.get_current_executions()
+        self._exec_queue = [e for e in self._exec_queue if e in exec_queue]
+        if self._exec_queue:
+            self.schedule_update_ha_state(True)
+            return
+
         self.controller.get_states([self.tahoma_device])
         self.update_temp(None)
         if self._widget == W_ST:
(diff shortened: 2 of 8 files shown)


vlebourl:
If accepted, I think this would deserve a beta release.

custom_components/tahoma/binary_sensor.py
```diff
@@ -60,6 +60,12 @@ def device_class(self):
 
     def update(self):
         """Update the state."""
+        exec_queue = self.controller.get_current_executions()
```
iMicknl, line 63: This code is the same in all implementations? Should we move it to the `tahoma_device` and call this function from here?

vlebourl: It could take this form:
```
    def should_wait(self):
        """Wait for actions to finish (this would be in tahoma_device"""
        exec_queue = self.controller.get_current_executions()
        self._exec_queue = [e for e in self._exec_queue if e in exec_queue]
        return True if self._exec_queue else False

    def update(self):
        """Update the state."""
        if self.should_wait():
            self.schedule_update_ha_state(True)
            return
        # continue update
```

rabesocke:
I test with the 1.5-alpha2. My covers reacts with several seconds delay to the click in the HA frontend.

vlebourl:
How reactive is it when interacting through the TaHoma web dashboard? 
Also, when clicking in the HA frontend, can you keep the Tahomalink website open and visible, you should get something like this: 
image
Is the message on TaHoma dashboard appearing with the same delay?

vlebourl:
It's very reactive at my place... Not sure why it would add any delay.

rabesocke:
There is no delay through the TaHoma Dashobard. I can change the position, wait a little bit and click on stop and the cover reacts immediately. 

Through HA it starts in general directly but not always. When there is a delay then you see the delay also on the TaHoma dashboard.

One interesting point: The stop button is not working while moving. When I click the button on TaHoma dashboard it stops immediately. (Please wait some seconds after you start the movement and then try to click on stop - for me it is not working)

vlebourl:
"When there is a delay then you see the delay also on the TaHoma dashboard."
Do you mean there is the delay for the message in my screen capture to appear? 
If so, the delay is in the communication between HA and TaHoma, not between TaHoma and the Cover.
Be aware that the api is an unofficial one, not intended by Somfy to be used by HA. Hence sending lots of requests in a row might get you kicked out, forcing a re-log that might take few seconds.

I don't know about the stop button, but I suspect it's not related to the delay, so I propose to keep it aside for now.

rabesocke:
The delay is completely new to me. Before I had to use the current version (because of the login issue) there were no such delays. It appears the first timt 30 minutes ago after the HA update to 0.112 and the component update to 1.5-alpha2

vlebourl:
The delay might be related to the login issue. I probably still get logged out, then the api re-logs you in automatically causing the delay. We would need to figure out why you get logged out, but I have no idea how to do that...

jurgen272:
Overhere it is far beter then in the last versions. Before the stop button wasn't reacting, and now it is. will test again in few hours.
0.112 en alpha2

rabesocke:
> The delay might be related to the login issue. You probably still get logged out a lot, then the api re-logs you in automatically causing the delay. We would need to figure out why you get logged out, but I have no idea how to do that...
Ok, that could be the reason for the delay. It would be great to get this fixed - but I also have no idea, how to find the reason

> Overhere it is far beter then in the last versions. Before the stop button wasn't reacting, and now it is. will test again in few hours.
Interesting ... for me all versions including 1.3 were much better than the current version and the stop button was working for me without problems. The problems for me start with version 1.4 (which I skipped until the login issue)

vlebourl:
Can you open a new issue with details on your cover regarding the stop button?

I'll try to put more log messages to find why you get kicked out of the api.

rabesocke:
I tested it again and I would say that the reason is the login issue. I clicked on the move-down-button and it had a delay of around 2 seconds. The same happens when I clicked on the stop-button some seconds later. So I would say it relates not only to the stop-button.

vlebourl:
can you edit the file `tahoma_api.py` and add the following line 157, just before the `if not self.login()`
```
_LOGGER.debug("not authenticated: login again.")
```
Restart HA a try again with debug log setup for the component ([see here](https://github.com/iMicknl/ha-tahoma#advanced))

vlebourl:
By the way, conversation should be moved to #86

PR #95, opened 2020-07-02
Re-log in the API after a "error":"Not authenticated"

vlebourl:
Add new api functions:
* `is_authenticated`
* `logout`

Add a relog in `send_request` if `is_authenticated` is `False` fixing issue #86


lindemar:
This implementation in the tahoma_api.py fixed my login issue after updating TaHoma Box to version 2020.3.4-12. Thanks a lot for your effort!

Issue #96, opened 2020-07-02
Update for light component (rgb) fails - error in logs


jurgen272:
**Describe the bug**
RGB light is working correct, but I still have some errors in the loggings.

Errors:

Logboekdetails (ERROR)
Logger: homeassistant.helpers.entity
Source: custom_components/tahoma/light.py:160
First occurred: 8:06:58 (327 occurrences)
Last logged: 10:58:22

Update for light.light_terras_rgb fails
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 272, in async_update_ha_state
    await self.async_device_update()
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 466, in async_device_update
    self.update  # type: ignore
  File "/usr/local/lib/python3.7/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/config/custom_components/tahoma/light.py", line 160, in update
    CORE_BLUE_COLOR_INTENSITY_STATE
TypeError: color_RGB_to_hs() missing 2 required positional arguments: 'iG' and 'iB'

**Environment (please complete the following information):**
 - **Home Assistant version:** 0.112.0

**Device:** (if your problem is related to a specific device)
 - **Type**: io:DimmableRGBLightIOCompo

vlebourl:
Hi, 
Thanks for the report. Can you edit file `light.py` (using File Editor add-on for instance) and change line 151 to the end with:
```
        if CORE_RED_COLOR_INTENSITY_STATE in self.tahoma_device.active_states:
            self._hs_color = color_util.color_RGB_to_hs(
                self.tahoma_device.active_states.get(
                    CORE_RED_COLOR_INTENSITY_STATE
                ),
                self.tahoma_device.active_states.get(
                    CORE_GREEN_COLOR_INTENSITY_STATE
                ),
                self.tahoma_device.active_states.get(
                    CORE_BLUE_COLOR_INTENSITY_STATE
                )
            )
```

Restart HA and let me know whether you still get the errors.
Thanks

jurgen272:
Hi,

I changed:

        if CORE_RED_COLOR_INTENSITY_STATE in self.tahoma_device.active_states:
            self._hs_color = color_util.color_RGB_to_hs(
                [
                    self.tahoma_device.active_states.get(
                        CORE_RED_COLOR_INTENSITY_STATE
                    ),
[...]

PR #97, opened 2020-07-02
[fix] color_RGB_to_hs is missing parameters

vlebourl:
Related Github issues: #96


vlebourl:
Tested and validated.

PR #106, opened 2020-07-05
Split update cover method

tetienne:
To ease the maintenance and future improvement split the huge update method in several smaller parts.

### DIFF
--- a/custom_components/tahoma/cover.py
+++ b/custom_components/tahoma/cover.py
@@ -56,12 +56,13 @@ async def async_setup_entry(hass, entry, async_add_entities):
 
     data = hass.data[DOMAIN][entry.entry_id]
 
-    entities = []
     controller = data.get("controller")
 
-    for device in data.get("devices"):
-        if TAHOMA_TYPES[device.uiclass] == "cover":
-            entities.append(TahomaCover(device, controller))
+    entities = [
+        TahomaCover(device, controller)
+        for device in data.get("devices")
+        if TAHOMA_TYPES[device.uiclass] == "cover"
+    ]
 
     async_add_entities(entities)
 
@@ -72,6 +73,9 @@ class TahomaCover(TahomaDevice, CoverEntity):
     def __init__(self, tahoma_device, controller):
         """Initialize the device."""
         super().__init__(tahoma_device, controller)
+
+        self._tilt_position = None
+        self._position = None
 
         self._icon = None
         self._lock_timer = 0  # Can be 0 and bigger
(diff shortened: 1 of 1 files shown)


tetienne:
If OK, I will also move the code from `update_position` and `update_tilt_position" to their respective properties. Their is no need to intermediate properties.

custom_components/tahoma/cover.py
```diff
@@ -95,38 +99,31 @@ def update(self):
 
         self.controller.get_states([self.tahoma_device])
 
-        # Set current position.
+        self.update_postion()
```
iMicknl, line 102:
```suggestion
        self.update_position()
```

Typo

custom_components/tahoma/cover.py
```diff
@@ -95,38 +99,31 @@ def update(self):
...
+        self.update_tilt_position()
+        self.update_lock()
+
+    def update_position(self):
+
         # Home Assistant: 0 is closed, 100 is fully open.
         # core:ClosureState: 100 is closed, 0 is fully open.
 
+        states = self.tahoma_device.active_states
```
iMicknl, line 111: Good one, I like this! Would be good to have this in all components, makes it more readable.

custom_components/tahoma/cover.py
```diff
@@ -95,38 +99,31 @@ def update(self):
...
-
-        # Set tilt position for slats
-        if CORE_SLATS_ORIENTATION_STATE in self.tahoma_device.active_states:
-            self._tilt_position = 100 - self.tahoma_device.active_states.get(
-                CORE_SLATS_ORIENTATION_STATE
-            )
+        if CORE_TARGET_CLOSURE_STATE in states:
+            self._position = 100 - states.get(CORE_TARGET_CLOSURE_STATE)
 
         if getattr(self, "_position", False):
```
iMicknl, line 128:
```suggestion
        if self._position == False:
```

Could this be simplified like the others? Or would it return `False` if it is None as well?

custom_components/tahoma/cover.py
tetienne: Yep, it something I like to do when you repeat a lot a long path.

custom_components/tahoma/cover.py
tetienne: Fixed.

custom_components/tahoma/cover.py
tetienne: I missed this one you are right. self._position can be only an integer or be None. The ` getattr(self, "_position", False)` was a trick (ugly IMO). 
Before this PR self._position was not always existing. Now I init it in the __init__ method. 
The `getattr(self, "_position", False)` was here to return False in the case where self._position was not existing. We don't need it anymore.

custom_components/tahoma/cover.py
```diff
@@ -211,39 +209,26 @@ def set_cover_tilt_position(self, **kwargs):
...
+            return states.get("core:OpenClosedPedestrianState") == "closed"
+
+        if "core:OpenClosedUnknownState" in states:
+            return states.get("core:OpenClosedUnknownState") == "closed"
+
+        if self._position is not None:
             return self._position == 0
 
-        if getattr(self, "_tilt_position", None) is not None:
+        if self._tilt_position is not None:
```
tetienne, line 231: To be checked, but IMO the is_closed is only related to the position not the tilt position.

tetienne:
I've updated the PR according to your remarks.

custom_components/tahoma/cover.py
iMicknl: I own a device (BioclimaticPergola) which only has slats that could be tilted (or closed). In that case only `tilt_position` exists.

custom_components/tahoma/cover.py
```diff
@@ -56,12 +56,13 @@ async def async_setup_entry(hass, entry, async_add_entities):
 
     data = hass.data[DOMAIN][entry.entry_id]
 
-    entities = []
     controller = data.get("controller")
 
-    for device in data.get("devices"):
-        if TAHOMA_TYPES[device.uiclass] == "cover":
-            entities.append(TahomaCover(device, controller))
+    entities = [
```
iMicknl, line 61: Would be good to port this change to all platforms as well, in a future PR.

custom_components/tahoma/cover.py
tetienne: OK I understand. In home assistant, it will ease its usage indeed. In the UI, there is no button to close or open the tilt? And their is no visual indication if your pergola is opened or closed without this trick?

custom_components/tahoma/cover.py
tetienne: I will.

custom_components/tahoma/cover.py
iMicknl: There is a button to open and close the tilt, see `open_cover_tilt` and `close_cover_tilt`. Based on SUPPORTED_FEATURES the required buttons are shown in HA.

iMicknl: Toggling SUPPORTED_FEATURES based on the available commands, this is very dynamic and supports multiple device types without any hardcoded list.

tetienne: How does it look like without this fix? Sorry for my question, but often due to little hack, it's hard to understand the code later.

iMicknl: It is not really a hack, it first does the check if it is closed based on the state. If the state is not available, it has a look at the current position. If the current position is unknown, it falls back to the tilt position.

PR #108, opened 2020-07-06
Remove lock variables from Cover device state attributes

tetienne:
Some `_lock_xxx` variables was only used fo within `device_state_attributes` which is IMO useless. Removing them, we once again ease the code with useless boilerplate.

### DIFF
--- a/custom_components/tahoma/cover.py
+++ b/custom_components/tahoma/cover.py
@@ -228,12 +228,12 @@ class TahomaCover(TahomaDevice, CoverEntity):
     @property
     def icon(self):
         """Return the icon to use in the frontend, if any."""
-        icon = None
         if self._lock_timer > 0:
-            icon = "mdi:lock-alert"
             if self._lock_originator == "wind":
-                icon = "mdi:weather-windy"
-        return icon
+                return "mdi:weather-windy"
+            else:
+                return "mdi:lock-alert"
+        return None
 
     def open_cover(self, **kwargs):
         """Open the cover."""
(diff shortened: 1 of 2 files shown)


iMicknl:
@tetienne I will have a look later today, this is something I will need to test a bit more. I am not 100% sure what the previous code did, since this is coming from the original component.

custom_components/tahoma/cover.py
```diff
@@ -269,13 +228,12 @@ def device_state_attributes(self):
...
         if self._lock_timer > 0:
+            icon = "mdi:lock-alert"
             if self._lock_originator == "wind":
-                return "mdi:weather-windy"
-            else:
-                return "mdi:lock-alert"
-
-        return self._icon
+                icon = "mdi:weather-windy"
+        return icon
```
iMicknl, line 236: I would prefer to have this without the icon variable, like we do in `binary_sensor` and `sensor`. Where we directly return the value that is appropriate. 

```
if self._lock_timer > 0:
   if self._lock_originator == "wind":
       return "mdi:weather-windy"

   return "mdi:lock-alert"

return None
```

tetienne: OK, I will keep in mind this rule. Fixed.

tetienne:
@iMicknl Normally the data I removed are not used in the cover behavior. They are here only to be used as HA attributes where you can put what you want.

vlebourl:
I don't have covers and can't really appreciate those changes, so I put my faith in you :)

iMicknl:
@tetienne true, but I am not sure if those value are used by automations of people using the current component. I have never used those values, thus I would need to see if this is something where people can depend on. 

However, I am up for making the cover platform simpler, since there is still some legacy inside.

tetienne:
I don't see any reason to use these weird variables. It's not something familiar within HA.

Issue #111, opened 2020-07-06
Retrieve core:MeasuredValueType to understand if it is Celsius or Kelvin


iMicknl:
**Is your feature request related to a problem? Please describe.**
Currently we only support temperature sensors in Celsius.

**Describe the solution you'd like**
Would be good to retrieve the `core:MeasuredValueType` to understand if a `TemperatureSensor` is using Kelvin or Celsius. In other integrations I have seen that this is calculated based on the figure, which is more error prone and less nice.

https://github.com/iMicknl/ha-tahoma/blob/dd3a5c67035e591429071439b818ca991dca6122/custom_components/tahoma/sensor.py#L77-L79

PR #116, opened 2020-07-06
Parse temperature unit for TemperatureSensor

vlebourl:
* Add attributes if any in the API.
* Parse Temperature sensor for unit.

Related Github issues: #111


custom_components/tahoma/tahoma_api.py
```diff
@@ -629,19 +634,24 @@ def label(self):
...
+        """List of state definition."""
+        return self.__states_def
 
     @property
     def active_states(self):
         """Get active states."""
         return self.__active_states
 
+    @property
+    def attributes(self):
```
iMicknl, line 651: Good to have those present! Would be good to understand if we can make the component more dynamic by looking at the attributes, for example for covers and sensors.

custom_components/tahoma/sensor.py
```diff
@@ -75,7 +78,15 @@ def unit_of_measurement(self):
         states = self.tahoma_device.active_states
 
         if CORE_TEMPERATURE_STATE in states:
-            # TODO Retrieve core:MeasuredValueType to understand if it is Celsius or Kelvin
+            unit = self.tahoma_device.attributes.get(CORE_MEASURED_VALUE_TYPE)
+            if unit:
```
iMicknl, line 82: Why not check for Celsius? Now we have a double fallback to Celsius, but it is not super readable in my opinion. I would like it more if it would be something like this.

(I still miss switch/case in Python)

```python
        if CORE_TEMPERATURE_STATE in states:
            measured_value = self.tahoma_device.attributes.get(CORE_MEASURED_VALUE_TYPE)

            if measured_value is not None:
                if "Celcius" in measured_value:
                    return TEMP_CELSIUS

                if "Kelvin" in measured_value:
                    return TEMP_KELVIN

                if "Fahrenheit" in measured_value:
                    return TEMP_FAHRENHEIT       
           
            return TEMP_CELSIUS
```

vlebourl: I just thought that way was more pythonesque, but I don't really mind TBH.

tetienne: This is indeed IMO not really readable. You can use a dict to avoid the if else.

iMicknl: Do we need to check if the value is in the string? Or can we do a string compare, would be even more readable imo.

```python
if "Celsius" in measured_value:
```

```python
if measured_value == "Celsius":
```

vlebourl: `core:TemperatureInCelcius`
I don't have access to Fahrenheit and Kelvin. Not even sure they exist, but I guess it would be the same kind of format.

vlebourl: do you like this better:
```
        if CORE_TEMPERATURE_STATE in states:
            units = {
                "core:TemeratureInCelsius": TEMP_CELSIUS,
                "core:TemeratureInKelvin": TEMP_KELVIN,
                "core:TemeratureInFahrenheit": TEMP_FAHRENHEIT,
            }
            return units.get(self.tahoma_device.attributes.get(CORE_MEASURED_VALUE_TYPE), TEMP_CELSIUS)
```

custom_components/tahoma/tahoma_api.py
```diff
@@ -614,6 +614,11 @@ def __init__(self, protocol, dataInput):
                     self.__definitions["states"].append(state["qualifiedName"])
             self.__command_def = definition.get("commands")
             self.__states_def = definition.get("states")
+        self.__attributes = {}
+        attributes = dataInput.get("attributes")
+        if attributes:
+            for attr in attributes:
+                self.__attributes[attr["name"]] = attr["value"]
```
tetienne, line 621:
```suggestion
self.__attributes = {attr["name"]: attr["value] for attr in attributes}
```

custom_components/tahoma/tahoma_api.py
```diff
@@ -629,19 +634,24 @@ def label(self):
...
 
     @property
     def active_states(self):
         """Get active states."""
         return self.__active_states
 
+    @property
+    def attributes(self):
+        """Return entity attributes if any."""
+        return self.__attributes
```
tetienne, line 653: Why create such variable `self.__attributes`? You can directly use self.attributes without creating this property method.

custom_components/tahoma/sensor.py
```diff
@@ -75,7 +78,18 @@ def unit_of_measurement(self):
         states = self.tahoma_device.active_states
 
         if CORE_TEMPERATURE_STATE in states:
-            # TODO Retrieve core:MeasuredValueType to understand if it is Celsius or Kelvin
+            measured_value = self.tahoma_device.attributes.get(CORE_MEASURED_VALUE_TYPE)
+
+            if measured_value is not None:
```
tetienne, line 83:
```suggestion
            if measured_value:
```

custom_components/tahoma/sensor.py
```diff
@@ -75,7 +78,18 @@ def unit_of_measurement(self):
...
+                if "Celsius" in measured_value:
```
tetienne, line 84: So measured_value is a string that contains something else than Celsius?

custom_components/tahoma/tahoma_api.py
vlebourl: mostly because that's how the API was written by philklei. We do plan to rewrite the whole stuff anyways, so I wouldn't spend too much time cleaning it up for now.

custom_components/tahoma/sensor.py
tetienne: Looks better indeed.

custom_components/tahoma/sensor.py
vlebourl: yes, in my case it `core:TemperatureInCelcius`

custom_components/tahoma/tahoma_api.py
tetienne: OK. Do you plan to contribute back to its repository or create your own?

vlebourl: Not sure. We already made quite some changes, and in the end are planning to have an async api instead of the sync one we have ATM.

custom_components/tahoma/tahoma_api.py
```diff
@@ -614,6 +614,9 @@ def __init__(self, protocol, dataInput):
                     self.__definitions["states"].append(state["qualifiedName"])
             self.__command_def = definition.get("commands")
             self.__states_def = definition.get("states")
+        self.__attributes = {}
```
tetienne, line 617: Become useless as you reassign it 2 lines after.

custom_components/tahoma/tahoma_api.py
iMicknl: @tetienne we are looking into rewriting the python api wrapper, where we start from scratch. If you are interested in contributing, let me know. We have an example in a separate repository. :-)

tetienne: For sure I'm interesting :) It sounds familiar to me :)

custom_components/tahoma/sensor.py
iMicknl: Is it Celcius or Celsius? :D (just to be sure, I make this mistake so often that I wouldn't be surprised if Somfy developers did the same)

vlebourl: damn you're write: 
image

custom_components/tahoma/tahoma_api.py
iMicknl: I will add you later today, we are in the beginning phase, so would love to learn from your learning of the Somfy component.

We are now in a stage where we understand what functionality of the TaHoma API is needed. The current modified version in this repository doesn't have a high code quality, and I was not looking into improving the current one tbh.

custom_components/tahoma/sensor.py
```diff
@@ -75,8 +78,14 @@ def unit_of_measurement(self):
         states = self.tahoma_device.active_states
 
         if CORE_TEMPERATURE_STATE in states:
-            # TODO Retrieve core:MeasuredValueType to understand if it is Celsius or Kelvin
-            return TEMP_CELSIUS
+            return {
```
iMicknl, line 81: I like it!

custom_components/tahoma/sensor.py
```diff
@@ -75,8 +78,14 @@ def unit_of_measurement(self):
...
+                "core:TemperatureInCelcius": TEMP_CELSIUS,
```
iMicknl, line 82: Should we add a note that they have a typo here? (and that it is not in our code). Or add the correctly spelled one as well, to be sure.

vlebourl: added the correct one just in case

PR #120, opened 2020-07-06
Remove obsolete YAML schema

### LINKED ISSUE #113: Remove YAML config schema
https://github.com/iMicknl/ha-tahoma/blob/dd3a5c67035e591429071439b818ca991dca6122/custom_components/tahoma/__init__.py#L23-L37

Integration still has some legacy YAML schema, which is not used anymore. 

Should be removed. (or support for YAML could be added, for backwards compatibility)

iMicknl:
Fixes #113

Currently we don't use the old YAML schema for configuring the integration. Do you think we need this in the integration to ease the migration? 

I would favour for a breaking change where we drop YAML support, which also happens for most other integrations.

### DIFF
--- a/custom_components/tahoma/__init__.py
+++ b/custom_components/tahoma/__init__.py
@@ -20,22 +20,6 @@ from .tahoma_api import TahomaApi
 
 _LOGGER = logging.getLogger(__name__)
 
-# TODO Deprecate EXCLUDE for the native method of disabling entities
-CONFIG_SCHEMA = vol.Schema(
-    {
-        DOMAIN: vol.Schema(
-            {
-                vol.Required(CONF_USERNAME): cv.string,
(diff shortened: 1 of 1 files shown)


tetienne:
It's ok for me.

vlebourl:
we can add an option to exclude devices from the list of discovered devices as an option flow

iMicknl:
> Hum, if it remove the exlusion process, it will be annoying.

That was my first thought as well, however this is something that is natively possible in Home Assistant right now. You can disable entities (and later on devices hopefully) in the entity registry now. 

https://developers.home-assistant.io/docs/entity_registry_disabled_by/

tetienne:
@iMicknl Indeed, so we don't need anymore this option.

iMicknl:
@tetienne but good to mention it as a breaking change, when we merge to core.

PR #121, opened 2020-07-06
Code readability improvements

iMicknl:


custom_components/tahoma/light.py
```diff
@@ -131,26 +134,29 @@ def effect(self) -> str:
...
-            self._brightness = self.tahoma_device.active_states.get(
-                "core:LightIntensityState"
-            )
+        states = self.tahoma_device.active_states
+
+        if CORE_LIGHT_INTENSITY_STATE in states:
+            self._brightness = states.get(CORE_LIGHT_INTENSITY_STATE)
 
-        if self.tahoma_device.active_states.get("core:OnOffState") == "on":
+        if states.get(CORE_ON_OFF_STATE) == "on":
```
tetienne, line 148:
```suggestion
self._state = states.get(CORE_ON_OFF_STATE) == "on":
```

tetienne: My remark was to replace the if AND else block.

iMicknl: @tetienne I understood, GitHub doesn't support suggestions on multiple lines yet... I have fixed it manually, good suggestion.

custom_components/tahoma/light.py
```diff
@@ -131,26 +134,27 @@ def effect(self) -> str:
...
 
-        if self.tahoma_device.active_states.get("core:OnOffState") == "on":
-            self._state = True
-        else:
-            self._state = False
+        if CORE_LIGHT_INTENSITY_STATE in states:
+            self._brightness = states.get(CORE_LIGHT_INTENSITY_STATE)
+
+        if CORE_ON_OFF_STATE in states:
+            self._state = states.get(CORE_ON_OFF_STATE) == "on"
```
tetienne, line 149: IMO, the condition are useless. Directly get the value and store them in the property. If not existing, _brightness will be None and _state False.

custom_components/tahoma/tahoma_device.py
```diff
@@ -67,20 +65,19 @@ def assumed_state(self):
...
         }
 
-        if CORE_RSSI_LEVEL_STATE in self.tahoma_device.active_states:
-            attr[ATTR_RSSI_LEVEL] = self.tahoma_device.active_states[
-                CORE_RSSI_LEVEL_STATE
-            ]
+        states = self.tahoma_device.active_states
+
+        if CORE_RSSI_LEVEL_STATE in states:
+            attr[ATTR_RSSI_LEVEL] = states.get(CORE_RSSI_LEVEL_STATE)
```
tetienne, line 77: Same remark than above.

iMicknl: This one is different, it will only set the rssi level attribute if it exists. It doesn't exist for all Somfy TaHoma devices, thus this should be a dynamic assignment.

tetienne: It would be interesting to see what HA do with NONE attribute. But yes, we can keep it like this.

PR #123, opened 2020-07-07
Add an auto changelog issue

vlebourl:
Pushing into master will trigger a new changelog in branch update-changelog.


.github/workflows/changelog.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+      - name: Generate changelog
+        uses: charmixer/auto-changelog-action@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Commit files
+        env:
+          CI_USER: iMicknl
+          CI_EMAIL: mick@imick.nl
```
tetienne, line 23: Need a secret here.

vlebourl: this info is public though, took it from the public github profile.

.github/workflows/changelog.yaml
```diff
@@ -19,8 +19,8 @@ jobs:
 
       - name: Commit files
         env:
-          CI_USER: iMicknl
-          CI_EMAIL: mick@imick.nl
+          CI_USER: ${{ secrets.YOUR_GITHUB_USER }}
```
iMicknl, line 22: Added both to secrets

Issue #110, opened 2020-07-06
Refactor constants and strings


iMicknl:
- [x] Move constants from `consts.py` to the platform files if they are only used there
- [x] Add constants for all strings, especially in `binary_sensor`, `cover` and `switch`
- [x] Utilise f strings instead of .format() ([core comment](https://github.com/home-assistant/core/pull/36258#issuecomment-637720333))
- [x] Add helper functions for retrieving state and calling commands, eliminating all the if statements

tetienne:
> Add helper functions for retrieving state and calling commands, eliminating all the if statements

IMO it must be done in another issue. Otherwise, it will increase the size of your PR and made difficult to review it.

vlebourl:
Second tick by #126

PR #126, opened 2020-07-07
Add constants for all strings

vlebourl:
Related Github issues: #110


vlebourl:
any idea why that validation is failing?

vlebourl:
i've run it on a couple of merged PR where it was successful, it fails as well. It looks like a bug on their side...

iMicknl:
> any idea why that validation is failing?

The error is `ModuleNotFoundError: No module named 'voluptuous'` in HACS validation, however I am not sure why this fails.

custom_components/tahoma/binary_sensor.py
```diff
@@ -89,13 +99,13 @@ def icon(self) -> Optional[str]:
...
+                return ICON_WATER
             else:
-                return "mdi:water-off"
+                return ICON_WATER
 
         icons = {
-            DEVICE_CLASS_GAS: "mdi:waves",
-            DEVICE_CLASS_RAIN: "mdi:weather-rainy",
+            DEVICE_CLASS_GAS: ICON_WAVES,
+            DEVICE_CLASS_RAIN: ICON_WATER_OFF,
```
iMicknl, line 108: `ICON_WATER_OFF` is another icon?

custom_components/tahoma/const.py
```diff
@@ -38,8 +38,9 @@
...
 }
 
-# TaHoma internal device states
 CORE_ON_OFF_STATE = "core:OnOffState"
 
-# Commands
 COMMAND_OFF = "off"
+COMMAND_ON = "on"
+
+ICON_WEATHER_WINDY = "mdi:weather-windy"
```
iMicknl, line 46: Tbh, if an ICON is used in two integrations I would just duplicate it for readability and manageability.

custom_components/tahoma/light.py
```diff
@@ -100,14 +105,14 @@ def turn_on(self, **kwargs) -> None:
 
         if ATTR_BRIGHTNESS in kwargs:
             self._brightness = int(float(kwargs[ATTR_BRIGHTNESS]) / 255 * 100)
-            self.apply_action("setIntensity", self._brightness)
+            self.apply_action(COMMAND_SET_INTENSITY, self._brightness)
 
         elif ATTR_EFFECT in kwargs:
             self._effect = kwargs[ATTR_EFFECT]
-            self.apply_action("wink", 100)
+            self.apply_action(COMMAND_WINK, 100)
```
iMicknl, line 112:
```suggestion
            self.apply_action(self._effect, 100)
```

This is something we should have a look at. Is this not something that is passed dynamically? Not sure if self._effect is really needed as well.

custom_components/tahoma/lock.py
```diff
@@ -46,29 +48,21 @@ def update(self):
             return
 
         self.controller.get_states([self.tahoma_device])
-        self._battery_level = self.tahoma_device.active_states["core:BatteryState"]
-        self._name = self.tahoma_device.active_states["core:NameState"]
         if (
-            self.tahoma_device.active_states.get("core:LockedUnlockedState")
+            self.tahoma_device.active_states.get(CORE_LOCKED_UNLOCKED_STATE)
             == STATE_LOCKED
```
iMicknl, line 53: `STATE_LOCKED` is a constant from Home Assistant, not from the TaHoma component. Do we want to use those? My preference would be to have a specific one from the component, to keep it away from HA core.

custom_components/tahoma/binary_sensor.py
```diff
@@ -89,13 +99,13 @@ def icon(self) -> Optional[str]:
         """Return the icon to use in the frontend, if any."""
         if self.device_class == DEVICE_CLASS_WATER:
             if self.is_on:
-                return "mdi:water"
+                return ICON_WATER
             else:
-                return "mdi:water-off"
+                return ICON_WATER
```
iMicknl, line 104:
```suggestion
                return ICON_WATER_OFF
```

custom_components/tahoma/binary_sensor.py
vlebourl: hmm not sure what I did there, probably a bad copy and paste

custom_components/tahoma/const.py
vlebourl: done

custom_components/tahoma/lock.py
vlebourl: done

custom_components/tahoma/binary_sensor.py
vlebourl: fixed

vlebourl:
That was all because of #128...

PR #128, opened 2020-07-08
Remove all self._* variables in cover module

tetienne:
All the self._* was useless and add a lot of noise.  I plan to remove them also on the other platforms. It will ease a lot the code.


tetienne:
Once all the platforms cleaned in the same way, the [update method](https://github.com/iMicknl/ha-tahoma/pull/128/files#diff-02d6902693bd9f8f03eede16d8ddfb9eR82) will be moved directly in TahomaDevice class.

iMicknl:
Looks good, will have a better look later. I believe this could also make it easier to implement the DataUpdateCoordinator at some point, where we use the event listener. :-)

PR #132, opened 2020-07-08
Modified changelog workflow

vlebourl:
from https://github.com/47degrees/scalacheck-toolbox/blob/master/.github/workflows/changelog.yml

also added release drafter from https://github.com/47degrees/scalacheck-toolbox/blob/master/.github/workflows/release-drafter.yml


.github/workflows/changelog.yaml
```diff
@@ -1,38 +1,68 @@
...
+            },
+            "bug": {
+              "prefix": "🐛 **Bug Fixes**",
+              "labels": ["bug"]
+            },
+            "dependency": {
+              "prefix": "📈 **Dependency updates**",
+              "labels": ["dependency-update", "scala-steward"]
+            }
+          }'
```
tetienne, line 59: So we will have to use these labels. We need to create them. We can also added them automatically based on the name of the PR.

vlebourl: I think we don't need these ones actually, I should remove them. 
We can do the auto labeling based on PR name with this: https://github.com/TimonVS/pr-labeler-action

.github/pr-labeler.yml
```diff
@@ -0,0 +1,4 @@
+enhancement: ['feature/*', 'feat/*', 'enhancement/*']
```
tetienne, line 1: I would make an other label for feature. Like this in the changelog you will not see the new feature mixed with the refactors for instance.

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,64 @@
...
+              "prefix": "🚀 **Features**",
+              "labels": ["enhancement"]
+            },
+            "documentation": {
+              "prefix": "📘 **Documentation**",
+              "labels": ["documentation"]
+            },
+            "bug": {
+              "prefix": "🐛 **Bug Fixes**",
+              "labels": ["bug"]
```
tetienne, line 53: + feature

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,64 @@
...
+            },
+          }'
+      - name: Create Changelog Pull Request
+        uses: peter-evans/create-pull-request@v2
+        with:
+          token: ${{ secrets.ADMIN_GITHUB_TOKEN }}
+          commit-message: Update CHANGELOG.md [ci skip]
```
tetienne, line 60: I think you can removed the [ci skip]

vlebourl: done

.github/workflows/changelog.yml
vlebourl: done

.github/pr-labeler.yml
vlebourl: done

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,68 @@
...
+            },
+            "bug": {
+              "prefix": "🐛 **Bug Fixes**",
+              "labels": ["bug"]
+            },
+          }'
+      - name: Create Changelog Pull Request
+        uses: peter-evans/create-pull-request@v2
+        with:
+          token: ${{ secrets.ADMIN_GITHUB_TOKEN }}
```
tetienne, line 63: @iMicknl Do we have this token?

iMicknl: No, what is needed here?

vlebourl: from this https://github.com/peter-evans/create-pull-request I think we can use the defaut GITHUB_TOKEN, right?

tetienne: I think so

PR #136, opened 2020-07-09
Fix light update failing.

vlebourl:
Light update fails if some features aren't available.

### DIFF
--- a/custom_components/tahoma/light.py
+++ b/custom_components/tahoma/light.py
@@ -152,4 +152,6 @@ class TahomaLight(TahomaDevice, LightEntity):
             states.get(CORE_GREEN_COLOR_INTENSITY_STATE),
             states.get(CORE_BLUE_COLOR_INTENSITY_STATE),
         ]
-        self._hs_color = None if r is None else color_util.color_RGB_to_hs(r, g, b)
+        self._hs_color = (
+            None if None in [r, g, b] else color_util.color_RGB_to_hs(r, g, b)
+        )
--- a/custom_components/tahoma/light.py
+++ b/custom_components/tahoma/light.py
@@ -144,10 +144,13 @@ class TahomaLight(TahomaDevice, LightEntity):
         states = self.tahoma_device.active_states
 
         self._state = states.get(CORE_ON_OFF_STATE) == STATE_ON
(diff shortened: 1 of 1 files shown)


custom_components/tahoma/light.py
```diff
@@ -144,10 +144,13 @@ def update(self):
...
         self._state = states.get(CORE_ON_OFF_STATE) == STATE_ON
-        self._brightness = states.get(CORE_LIGHT_INTENSITY_STATE)
 
-        self._hs_color = color_util.color_RGB_to_hs(
-            states.get(CORE_RED_COLOR_INTENSITY_STATE),
-            states.get(CORE_GREEN_COLOR_INTENSITY_STATE),
-            states.get(CORE_BLUE_COLOR_INTENSITY_STATE),
-        )
+        if CORE_LIGHT_INTENSITY_STATE in states:
+            self._brightness = states.get(CORE_LIGHT_INTENSITY_STATE)
```
iMicknl, line 149: What is the error you get? Because this should return None, which is the default value

cc @tetienne

vlebourl: yes but it goes through `color_util.color_RGB_to_hs` which doesn't accept NoneType.

iMicknl: Ok, makes sense, in that case you don't need this one for brightness, only for the hs_color.

custom_components/tahoma/light.py
```diff
@@ -144,10 +144,13 @@ def update(self):
...
+
+        if CORE_RED_COLOR_INTENSITY_STATE in states:
```
iMicknl, line 151: Good spot, however I would check for all 3 values, since none of them can return None.

vlebourl:
```
[r, g, b] = [
            states.get(CORE_RED_COLOR_INTENSITY_STATE),
            states.get(CORE_GREEN_COLOR_INTENSITY_STATE),
            states.get(CORE_BLUE_COLOR_INTENSITY_STATE),
        ]
self._hs_color = (
            None if None in [r, g, b] else color_util.color_RGB_to_hs(r, g, b)
        )
```
this is better I think

vlebourl: I also removed the test from brightness.

PR #138, opened 2020-07-09
fix cover set up

vlebourl:
AttributeError: 'TahomaCover' object has no attribute '_lock_originator'

### DIFF
--- a/custom_components/tahoma/cover.py
+++ b/custom_components/tahoma/cover.py
@@ -226,9 +226,11 @@ class TahomaCover(TahomaDevice, CoverEntity):
             attr[ATTR_MEM_POS] = self.tahoma_device.active_states[
                 CORE_MEMORIZED_1_POSITION_STATE
             ]
-        if self._lock_originator is not None:
-            attr[ATTR_LOCK_ORIG] = self._lock_originator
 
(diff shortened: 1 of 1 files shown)


custom_components/tahoma/cover.py
```diff
@@ -226,9 +226,11 @@ def device_state_attributes(self):
             attr[ATTR_MEM_POS] = self.tahoma_device.active_states[
                 CORE_MEMORIZED_1_POSITION_STATE
             ]
-        if self._lock_originator is not None:
-            attr[ATTR_LOCK_ORIG] = self._lock_originator
 
+        if IO_PRIORITY_LOCK_ORIGINATOR_STATE in self.tahoma_device.active_states:
+            attr[ATTR_LOCK_ORIG] = self.tahoma_device.active_states[
+                IO_PRIORITY_LOCK_ORIGINATOR_STATE
+            ]
```
tetienne, line 233: Oh Sorry I missed this one.

tetienne: Create a variable `state = self.tahoma_device.active_states` to avoid long lines.

vlebourl: damn... seen that too late...

PR #140, opened 2020-07-09
fix changlog workflow trigger.

vlebourl:


vlebourl:
This should trigger on every new tag, not just master releases. Is that fine with you?

vlebourl:
would you rather have us use this:
`v[0-9]+.[0-9]+.[0-9]+`

PR #141, opened 2020-07-09
Only trigger update if the device is available

vlebourl:
This prevents following messages in the log at startup:
`Unable to update from sensor: could not convert string to float: 'unavailable'`

### DIFF
--- a/custom_components/tahoma/tahoma_device.py
+++ b/custom_components/tahoma/tahoma_device.py
@@ -124,6 +124,8 @@ class TahomaDevice(Entity):
 
     def should_wait(self):
         """Wait for actions to finish."""
+        if not self.available:
+            return True
         exec_queue = self.controller.get_current_executions()
         self._exec_queue = [e for e in self._exec_queue if e in exec_queue]
         return True if self._exec_queue else False


iMicknl:
I am not sure why this should be in the should_wait() function? Where is that error message coming from.

vlebourl:
I think it can apply to all devices. When starting up HA, it loads entities in an un predictable order, and sometimes entities are loaded but still `unavailable`, in which case the update method fails. In any case, we should not try to update an entity that is not available. Putting it in should_wait allows to write it only once, fewer changes. I guess this will move when all update methods are moved in a single one in tahoma_device.

iMicknl:
I don't think this is the best approach. What if the state is unavailable based on the Tahoma state? (which happens if a device is out of reach for example). Looking at this statement, it looks like a device will never be updated to available since that is coming from the state, which won't be updated anymore for 'unavailable' devices?

vlebourl:
Indeed, didn't see that.

vlebourl:
I'll close this one as this is not a good approach, the underlying issue (that message popping up in the log) only appear at HA startup, and has no consequences when HA is fully started.

PR #143, opened 2020-07-10
Remove .format().

vlebourl:
States are already formatted as numbers in TahomaApi
`def set_active_state(self, name, value):` (line 652)

https://github.com/iMicknl/ha-tahoma/blob/a518ba3f9909a6713b0b1a2c0bb9be7a9bc5ac59/custom_components/tahoma/tahoma_api.py#L652

Github issue #110


tetienne:
Your branch is named f-string, but you didn't use f-string :D

vlebourl:
I know... I named the branch from the issue, but realized we didn't need the f strings...

PR #145, opened 2020-07-11
Remove all private properties

tetienne:


custom_components/tahoma/lock.py
```diff
@@ -46,49 +43,44 @@ def update(self):
...
         """Lock method."""
-        _LOGGER.debug("Locking %s", self._name)
+        _LOGGER.debug("Locking %s", self.name)
         self.apply_action("lock")
 
     @property
     def name(self):
         """Return the name of the lock."""
-        return self._name
+        return self.tahoma_device.active_states["core:NameState"]
```
iMicknl, line 60: def name() could be inherited from tahoma_device. NameState is also not the correct one to use, since this could be truncated for longer labels. Correct one is to use `label`.

custom_components/tahoma/lock.py
```diff
@@ -46,49 +43,44 @@ def update(self):
...
 
     @property
     def available(self):
         """Return True if the lock is available."""
-        return self._available
+        return (
+            self.tahoma_device.active_states.get("core:AvailabilityState")
+            == "available"
```
iMicknl, line 67: `available()` should be inherited from tahoma_device

custom_components/tahoma/lock.py
```diff
@@ -46,49 +43,44 @@ def update(self):
...
+        return (
+            self.tahoma_device.active_states.get("core:LockedUnlockedState")
+            == STATE_LOCKED
+        )
 
     @property
     def device_state_attributes(self):
         """Return the lock state attributes."""
-        attr = {ATTR_BATTERY_LEVEL: self._battery_level}
+        attr = {
```
iMicknl, line 81: `device_state_attributes` is not necessary, since `ATTR_BATTERY_LEVEL` will be added by `tahoma_device`

custom_components/tahoma/binary_sensor.py
```diff
@@ -64,16 +64,22 @@ async def async_setup_entry(hass, entry, async_add_entities):
...
-
     @property
     def is_on(self):
         """Return the state of the sensor."""
-        return self._state == STATE_ON
+
+        states = self.tahoma_device.active_states
+
+        return (
+            states.get(CORE_CONTACT_STATE) == "open"
```
iMicknl, line 74: Nice one! Great to have the active_states standardized, so we can move to a single update() method in `tahoma_device.py`

tetienne: Yep, it will remove a lot of boilerplate.

custom_components/tahoma/lock.py
tetienne: core:AvailabilityState exists for all the devices?

custom_components/tahoma/lock.py
tetienne: I prefer to do it in another PR to not increase the size of this one.

custom_components/tahoma/lock.py
vlebourl: I don't think so. That's why I left this one in lock.py

iMicknl: I believe it exists for more devices, since it exists in `tahoma_device`. I would need to check.

custom_components/tahoma/lock.py
iMicknl: Also fine, however instead of changing this part, you could just delete the whole `device_state_attributes`.

tetienne:
I still have to clean the climate platform. I keep it for the end...

custom_components/tahoma/lock.py
tetienne: That's the same check that the one in TahomaDevice. I've removed this one.

custom_components/tahoma/lock.py
tetienne: Done.

custom_components/tahoma/lock.py
tetienne: Done.

tetienne:
Climate is really too complex, I've only remove some unused variables.

tetienne:
Ready to be merged.

vlebourl:
> Climate is really too complex, I've only remove some unused variables.

I wrote that! It was tough to be honeset. I'll have a look later this week.

PR #147, opened 2020-07-12
Add cycle support for the cover platform

### LINKED ISSUE #146: io:CyclicGarageOpenerIOComponent not accessible by services, automations, and has status unknown
**Describe the bug**
I have installed the following Somfy [device](Physical device: https://www.somfy.fr/produits/1841229/recepteur-acces-compatible-io) which allows opening and closing of garage door via dry contact. Even if this is IO device, state cannot be fed back because the interface to the garage door motor is via dry contact (up/stop/down/stop/etc).

The device is recognised as a garage door opener and created in HA, but no actions (scropts/automations/status) are possible.

**To Reproduce**
1. Go to '.Configuration / Integration / Tahoma Integration Devices '
2. Click on 'right_garage_door'
3. Click on the plus sign next to "automations"
4. Message:
"Create automation with device
There are no automations available for this device."

**Expected behaviour**
Ability to create automations like with any other Tahoma component. 

**Log entry**
Logger: homeassistant.components.cover.reproduce_state
Source: components/cover/reproduce_state.py:51
Integration: Cover (documentation, issues)
First occurred: 17:06:03 (1 occurrences)
Last logged: 17:06:03
Invalid state specified for cover.right_garage_door: unknown

**Environment (please complete the following information):**
 - **Home Assistant version:** 0.112.4
 - **Platform**: Tahoma

**Device:** Cyclic Garage Door
 - **Model**: dry contact remote, enables to toggle existing garage door opener
 - **Type**: [e.g. `io:DimmableLightIOComponent`, can be gathered from device page in HA looking at Firmware] `/config/devices/dashboard`]
 - **Device details**

```json
{
   "creationTime":1594134595000,
   "lastUpdateTime":1594134595000,
   "label":"Left Garage Door",
   "deviceURL":"###",
   "shortcut":false,
   "controllableName":"io:CyclicGarageOpenerIOComponent",
   "definition":{
      "commands":[
         {
            "commandName":"cycle",
            "nparams":0
         },
         {
            "commandName":"delayedStopIdentify",
            "nparams":1
         },
         {
            "commandName":"getName",
            "nparams":0
         },
         {
            "commandName":"identify",
            "nparams":0
         },
         {
            "commandName":"setName",
            "nparams":1
         },
         {
            "commandName":"startIdentify",
            "nparams":0
         },
         {
            "commandName":"stopIdentify",
            "nparams":0
         },
         {
            "commandName":"wink",
            "nparams":1
         },
         {
            "commandName":"pairOneWayController",
            "nparams":2
         },
         {
            "commandName":"unpairAllOneWayControllers",
            "nparams":0
         },
         {
            "commandName":"unpairOneWayController",
            "nparams":2
         }
      ],
      "states":[
         {
            "values":[
               "good",
               "low",
               "normal",
               "verylow"
            ],
            "type":"DiscreteState",
            "qualifiedName":"core:DiscreteRSSILevelState"
         },
         {
            "type":"DataState",
            "qualifiedName":"core:NameState"
         },
         {
            "type":"ContinuousState",
            "qualifiedName":"core:PriorityLockTimerState"
         },
         {
            "type":"ContinuousState",
            "qualifiedName":"core:RSSILevelState"
         },
         {
            "values":[
               "available",
               "unavailable"
            ],
            "type":"DiscreteState",
            "qualifiedName":"core:StatusState"
         },
         {
            "values":[
               "comfortLevel1",
               "comfortLevel2",
               "comfortLevel3",
               "comfortLevel4",
               "environmentProtection",
               "humanProtection",
               "userLevel1",
               "userLevel2"
            ],
            "type":"DiscreteState",
            "qualifiedName":"io:PriorityLockLevelState"
         },
         {
            "values":[
               "LSC",
               "SAAC",
               "SFC",
               "UPS",
               "externalGateway",
               "localUser",
               "myself",
               "rain",
               "security",
               "temperature",
               "timer",
               "user",
               "wind"
            ],
            "type":"DiscreteState",
            "qualifiedName":"io:PriorityLockOriginatorState"
         }
      ],
      "dataProperties":[
         {
            "value":"500",
            "qualifiedName":"core:identifyInterval"
         }
      ],
      "widgetName":"CyclicGarageDoor",
      "uiClass":"GarageDoor",
      "qualifiedName":"io:CyclicGarageOpenerIOComponent",
      "type":"ACTUATOR"
   },
   "states":[
      {
         "name":"core:NameState",
         "type":3,
         "value":"Left Garage Door"
      },
      {
         "name":"core:PriorityLockTimerState",
         "type":1,
         "value":0
      },
      {
         "name":"core:StatusState",
         "type":3,
         "value":"available"
      },
      {
         "name":"core:DiscreteRSSILevelState",
         "type":3,
         "value":"good"
      },
      {
         "name":"core:RSSILevelState",
         "type":2,
         "value":84.0
      }
   ],
   "available":true,
   "enabled":true,
   "placeOID":"###",
   "widget":"CyclicGarageDoor",
   "type":1,
   "oid":"###",
   "uiClass":"GarageDoor"
}
```

**Additional context**

tetienne:
Fix #146

### DIFF
--- a/custom_components/tahoma/cover.py
+++ b/custom_components/tahoma/cover.py
@@ -19,10 +19,11 @@ from homeassistant.components.cover import (
     SUPPORT_SET_TILT_POSITION,
     SUPPORT_STOP,
     SUPPORT_STOP_TILT,
-    CoverEntity,
+    CoverDevice,
 )
 
 from .const import DOMAIN, TAHOMA_TYPES
+from .switch import COMMAND_CYCLE
 from .tahoma_device import TahomaDevice
 
 _LOGGER = logging.getLogger(__name__)
@@ -95,7 +96,7 @@ async def async_setup_entry(hass, entry, async_add_entities):
     async_add_entities(entities)
 
 
-class TahomaCover(TahomaDevice, CoverEntity):
+class TahomaCover(TahomaDevice, CoverDevice):
     """Representation a TaHoma Cover."""
 
     @property
@@ -254,3 +255,10 @@ class TahomaCover(TahomaDevice, CoverEntity):
             supported_features |= SUPPORT_CLOSE
 
         return supported_features
+
+    def toggle(self):
+        """Toggle the entity."""
+        if COMMAND_CYCLE in self.tahoma_device.command_definitions:
+            return self.apply_action(COMMAND_CYCLE)
+        else:
+            super().toggle()
--- a/custom_components/tahoma/cover.py
+++ b/custom_components/tahoma/cover.py
@@ -19,7 +19,7 @@ from homeassistant.components.cover import (
     SUPPORT_SET_TILT_POSITION,
     SUPPORT_STOP,
     SUPPORT_STOP_TILT,
-    CoverDevice,
+    CoverEntity,
 )
 
 from .const import DOMAIN, TAHOMA_TYPES
@@ -96,7 +96,7 @@ async def async_setup_entry(hass, entry, async_add_entities):
     async_add_entities(entities)
 
 
-class TahomaCover(TahomaDevice, CoverDevice):
+class TahomaCover(TahomaDevice, CoverEntity):
     """Representation a TaHoma Cover."""
 
     @property
@@ -258,7 +258,7 @@ class TahomaCover(TahomaDevice, CoverDevice):
 
     def toggle(self):
         """Toggle the entity."""
-        if COMMAND_CYCLE in self.tahoma_device.command_definitions:
-            return self.apply_action(COMMAND_CYCLE)
+        if self.has_command(COMMAND_CYCLE):
+            self.apply_action(COMMAND_CYCLE)
         else:
             super().toggle()


custom_components/tahoma/cover.py
```diff
@@ -19,10 +19,11 @@
     SUPPORT_SET_TILT_POSITION,
     SUPPORT_STOP,
     SUPPORT_STOP_TILT,
-    CoverEntity,
+    CoverDevice,
```
vlebourl, line 22: Isn't `*Device` deprecated in favor of `*Entity`

`CoverDevice is deprecated, modify to extend CoverEntity
11:00:49 AM – Cover (WARNING)`

iMicknl:
This will only enable the `cover.toggle()` service for users of this component, it won't update the interface for them. For advanced users this could be fine, however I prefer to have all devices supported in the UI as well.

We would need to understand if we could work around this, for example by setting assumed_state() to True + other ways to just show the toggle. 

A workaround could be to check for `cycle` in OPEN and CLOSE, as the last value. Thus supporting the buttons as a fallback.

custom_components/tahoma/cover.py
tetienne: Ah yes! But it's weird my ide only find CoverDevice not CoverEntity

tetienne:
I was just right now looking information to understand the assumed_state property. 
Do you know if true the ui display both open and close buttons?

iMicknl:
> I was just right now looking information to understand the assumed_state property.
> Do you know if true the ui display both open and close buttons?
I don't know to be honest. For lights it does show both buttons, but those are both mapped to a different command (on / off). However, since there is no state, you will need both buttons.

> IMO it will be confusing to have two buttons doing the same action.
Agree, however since there is no state, I don't think there is any other option. This is the behaviour for lights / switches in core already. The only thing that I don't really like is that on the background we are utilising the same command (`cycle`) twice.

However, having it only available via a service, and not via the UI, makes it even more confusing I am afraid. Best would be if there would just be a possibility to add `SUPPORTS_TOGGLE` to SUPPORTED_FEATURES.

custom_components/tahoma/cover.py
```diff
@@ -347,3 +348,10 @@ def supported_features(self):
             supported_features |= SUPPORT_CLOSE
 
         return supported_features
+
+    def toggle(self):
+        """Toggle the entity."""
+        if COMMAND_CYCLE in self.tahoma_device.command_definitions:
```
vlebourl, line 354: use select_command

tetienne:
@vlebourl Ready to be merged.

tetienne:
@iMicknl What do you think about creating an issue in the architecture repo about the SUPPORTS_TOGGLE?

iMicknl:
@tetienne I see this one is merged, but I don't believe it will work. Since SUPPORTS_OPEN / SUPPORTS_CLOSE is not here, the toggle function won't be enabled.

He can now toggle it via a service call, but in the interface there won't be any possibility to control. I would prefer to keep the issue topic open.

PR #150, opened 2020-07-16
Enhancement/add missing uiclasses

vlebourl:


custom_components/tahoma/__init__.py
```diff
@@ -59,7 +59,10 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
     for device in devices:
         _device = controller.get_device(device)
 
-        if _device.uiclass in TAHOMA_TYPES:
+        if (
+            _device.uiclass in TAHOMA_TYPES
+            and TAHOMA_TYPES[_device.uiclass] is not None
+        ):
```
tetienne, line 65: `if TAHOMA_TYPES.get(_device.uiclass])` is enough.

iMicknl:
@vlebourl I will close this one for now. Lets see if we need to add those in the future :). But I think we should just add classes that are supported, where other classes will show a debug error.

PR #152, opened 2020-07-16
Group entities that belong to the same device

tetienne:
Some of my device urls look like: "io://xxxxx/4468654#10", "io://xxxxx/4468654#9"... 
We can use this to group several entities in the same device.

### DIFF
--- a/custom_components/tahoma/tahoma_device.py
+++ b/custom_components/tahoma/tahoma_device.py
@@ -119,7 +119,7 @@ class TahomaDevice(Entity):
     def device_info(self):
         """Return device registry information for this entity."""
         return {
-            "identifiers": {(DOMAIN, self.unique_id)},
+            "identifiers": {(DOMAIN, self.unique_id.split("#", 1)[0])},
             "manufacturer": "Somfy",
             "name": self.name,
             "model": self.tahoma_device.widget,


custom_components/tahoma/tahoma_device.py
```diff
@@ -119,7 +119,7 @@ def device_state_attributes(self):
     def device_info(self):
         """Return device registry information for this entity."""
         return {
-            "identifiers": {(DOMAIN, self.unique_id)},
+            "identifiers": {(DOMAIN, self.unique_id.split("#", 1)[0])},
             "manufacturer": "Somfy",
             "name": self.name,
             "model": self.tahoma_device.widget,
```
vlebourl, line 125: name and model should not stay I guess, at least in the Thermostat, they're not the same for all 3 entities.

tetienne:
>>> unique_id="io://xxxxx/4468654#10"
>>> unique_id.split("#", 1)[0]
'io://xxxxx/4468654'
>>> unique_id = "foo"
>>> unique_id.split("#", 1)[0]
'foo'

vlebourl:
This is rejected as not possible in a satisfactory as Somfy doesn't have the Device abstraction as HA does.

Issue #151, opened 2020-07-16
Simplify climate integration


vlebourl:
ok, here are my thoughts regarding the climate entity:
- as it is now, I don't think it will be accepted in core
- the option flow is not very self explanatory
- Somfy has different logics depending on the devices (I know only of the Smart thermostat and AEH or now, but according to the uiClasses list, there are others)

What we can do:
- remove the option flow
- build the climate entity for smart thermostat by retrieving the temperature sensor with the same url, supporting target temperature and preset modes
- build a dumb climate entity only supporting presets for AEH (and maybe comparable devices we don't know of yet)

vlebourl:
Closed by #159

PR #159, opened 2020-07-20
Enhancement/refactor climate

vlebourl:
Refactor climate component.

* **Breaking change** Removed option flow as is no longer needed.
* Simplify AtlanticElectricalHeater component to support only preset modes to stick to what the physical device actually does.
  **Breaking change** This entity no longer supports target temperature.

* Separate devices into different classes to make the code more readable. 
* Add a base class for common properties.

Github Issue: #151


custom_components/tahoma/climate.py
```diff
@@ -67,179 +179,63 @@
...
-
-    def __init__(self, tahoma_device, controller, sensor_id=None, preset_temp=None):
-        """Initialize the sensor."""
-        super().__init__(tahoma_device, controller)
-        if preset_temp is None:
-            preset_temp = {}
-        self._temp_sensor_entity_id = sensor_id
-        self._current_temperature = 0
-        self._target_temp = None
+class SomfyThermostat(BaseClimateEntity):
```
tetienne, line 203: Can you move these new classes to a dedicated file? It will avoid having a big module.

custom_components/tahoma/climate.py
```diff
@@ -1,58 +1,170 @@
...
+        """Return a list of available preset modes."""
+        return self._preset_modes
+
+
+COMMAND_SET_HEATING_LEVEL = "setHeatingLevel"
+
+IO_TARGET_HEATING_LEVEL_STATE = "io:TargetHeatingLevelState"
+
+PRESET_FROST_PROTECTION = "frostprotection"
+PRESET_OFF = "off"
```
tetienne, line 114: Constant are usually at the top of the file.

custom_components/tahoma/climate.py
vlebourl: to be honest, I tried but it failed miserably... Somehow I couldn't get HA to start and correctly import the classes...

custom_components/tahoma/climate.py
tetienne: Oh, that's linked to the class below.

custom_components/tahoma/climate.py
```diff
@@ -1,58 +1,170 @@
...
+class BaseClimateEntity(TahomaDevice, ClimateEntity):
+    """Base class for Tahoma climate entity."""
+
+    def __init__(self, device, controller):
+        """Init method."""
+        super().__init__(device, controller)
+        self._preset_modes = None
+        self._preset_mode = None
+        self._hvac_modes = None
+        self._hvac_mode = None
```
tetienne, line 75: This cannot be retrieved from the states?

custom_components/tahoma/climate.py
vlebourl: done

custom_components/tahoma/climate.py
```diff
@@ -1,498 +1,34 @@
...
+    for device in climate_devices:
+        if device.widget == AEH:
+            entities.append(AtlanticElectricalHeater(device, controller))
+        elif device.widget == ST:
+            base_url = device.url.split("#", 1)[0]
+            sensor = None
+            for k, v in hass.data["entity_registry"].entities.items():
+                if v.unique_id == base_url + "#2":
+                    sensor = k
+                    break
```
tetienne, line 32: It seems you can retrieve an entity directly using its id:
```
entity_registry = await self.hass.helpers.entity_registry.async_get_registry()
entity_entry = entity_registry.async_get(f"{base_url}#2")
```

custom_components/tahoma/climate_st.py
```diff
@@ -0,0 +1,242 @@
...
+    STATE_PRESET_FREEZE: PRESET_FREEZE,
+    STATE_PRESET_MANUAL: PRESET_NONE,
+    STATE_PRESET_SLEEPING_MODE: PRESET_NIGHT,
+}
+PRESET_TEMPERATURES = {
+    PRESET_HOME: "somfythermostat:AtHomeTargetTemperatureState",
+    PRESET_AWAY: "somfythermostat:AwayModeTargetTemperatureState",
+    PRESET_FREEZE: "somfythermostat:FreezeModeTargetTemperatureState",
+    PRESET_NIGHT: "somfythermostat:SleepingModeTargetTemperatureState",
+}
```
tetienne, line 70: Nice

custom_components/tahoma/climate_st.py
```diff
@@ -0,0 +1,242 @@
...
+
+    def __init__(self, device, controller, sensor):
+        """Init method."""
+        super().__init__(device, controller)
+        self._temp_sensor_entity_id = sensor
+        self._saved_target_temp = (
+            self.select_state(PRESET_TEMPERATURES[self.preset_mode])
+            if self.hvac_mode == HVAC_MODE_AUTO
+            else self.select_state(CORE_DEROGATED_TARGET_TEMPERATURE_STATE)
+        )
```
tetienne, line 84: This is not really readable.

custom_components/tahoma/climate_st.py
```diff
@@ -0,0 +1,242 @@
...
+        def _async_startup(event):
+            """Init on startup."""
+            if self._temp_sensor_entity_id is not None:
+                temp_sensor_state = self.hass.states.get(self._temp_sensor_entity_id)
+                if temp_sensor_state and temp_sensor_state.state != STATE_UNKNOWN:
+                    self.update_temp(temp_sensor_state)
+
+        self.hass.bus.async_listen_once(EVENT_HOMEASSISTANT_START, _async_startup)
+
+        self.schedule_update_ha_state(True)
```
tetienne, line 104: It probably work, but I don't get it. What's the purpose?

custom_components/tahoma/climate_st.py
```diff
@@ -0,0 +1,242 @@
...
+        return [HVAC_MODE_AUTO, HVAC_MODE_HEAT]
+
+    @property
+    def preset_mode(self) -> Optional[str]:
+        """Return the current preset mode, e.g., home, away, temp."""
+        return MAP_PRESET_MODES[
+            self.select_state(ST_HEATING_MODE_STATE)
+            if self.hvac_mode == HVAC_MODE_AUTO
+            else self.select_state(ST_DEROGATION_HEATING_MODE_STATE)
+        ]
```
tetienne, line 151: Not really readable.

custom_components/tahoma/climate_st.py
```diff
@@ -0,0 +1,242 @@
...
+
+    @property
+    def current_temperature(self) -> Optional[float]:
+        """Return the current temperature."""
+        return self._current_temperature
+
+    @property
+    def min_temp(self) -> float:
+        """Return the minimum temperature."""
+        return 15.0
```
tetienne, line 172: Why?

custom_components/tahoma/climate_st.py
```diff
@@ -0,0 +1,242 @@
...
+        if temperature is None:
+            return
+
+        self.apply_action(
+            COMMAND_SET_DEROGATION, temperature, STATE_DEROGATION_FURTHER_NOTICE
+        )
+        self.apply_action(
+            COMMAND_SET_MODE_TEMPERATURE, STATE_PRESET_MANUAL, temperature
+        )
+        self.apply_action(COMMAND_REFRESH_STATE)
```
tetienne, line 200: Why refresh the state for this platform in particular?

custom_components/tahoma/climate_st.py
```diff
@@ -0,0 +1,242 @@
...
+            self.apply_action(COMMAND_EXIT_DEROGATION)
+            self.apply_action(COMMAND_REFRESH_STATE)
+        elif hvac_mode == HVAC_MODE_HEAT:
+            self.set_preset_mode(PRESET_NONE)
+
+    def set_preset_mode(self, preset_mode: str) -> None:
+        """Set new preset mode."""
+        if self.preset_mode == preset_mode:
+            return
+        modes = {
```
tetienne, line 217: Can be set as constant at the top of the module.

custom_components/tahoma/climate_st.py
```diff
@@ -0,0 +1,242 @@
...
+                COMMAND_SET_DEROGATION,
+                self._saved_target_temp,
+                STATE_DEROGATION_FURTHER_NOTICE,
+            )
+            self.apply_action(
+                COMMAND_SET_MODE_TEMPERATURE,
+                STATE_PRESET_MANUAL,
+                self._saved_target_temp,
+            )
+        self.apply_action(COMMAND_REFRESH_STATE)
```
tetienne, line 242: Same remark.

custom_components/tahoma/climate.py
vlebourl: done

custom_components/tahoma/climate_st.py
vlebourl: line 104? It's to first a first update at startup to get the temperature from the sensor freshly registered. I guess we can get rid of that line.

custom_components/tahoma/climate_st.py
vlebourl: is this better?
```
 @property
    def preset_mode(self) -> Optional[str]:
        """Return the current preset mode, e.g., home, away, temp."""
        st_preset = (
            self.select_state(ST_HEATING_MODE_STATE)
            if self.hvac_mode == HVAC_MODE_AUTO
            else self.select_state(ST_DEROGATION_HEATING_MODE_STATE)
        )
        return MAP_PRESET_MODES[st_preset]
```

custom_components/tahoma/climate_st.py
vlebourl: Because Somfy Smart Thermostat is limited to the range 15-26°C. As this property is part of the ClimateEntity, I think it's better to set that her, than check whether temperature is in the right range in `set_temperature` (although that should be done as well in case a user passes a temperature outside of the range in a service or an automation...)

custom_components/tahoma/climate_st.py
vlebourl:
```
COMMAND_REFRESH_DEROGATION = "refreshDerogation"
COMMAND_REFRESH_MODE_SETTING = "refreshModeSettings"
COMMAND_REFRESH_STATE = "refreshState"
```
these commands are part of ST but not AEH. To be fair, I'm not sure what it does, but at some point, TaHoma wasn't updating the states on its side without triggering the refreshState.
I'll try without it and see whether it should be kept.

custom_components/tahoma/climate_st.py
vlebourl: it can, but as it's only used there, I thought it would be clearer. I'll move it.

custom_components/tahoma/climate_st.py
vlebourl: same answer

custom_components/tahoma/climate_st.py
vlebourl: After some testing, the `refreshState` is needed to have the new state reflected in TaHoma, thus in HA.
plus opening the thermostat popup in tahomalink and sending a command triggers the refresh as well.
image

custom_components/tahoma/climate_st.py
tetienne: The whole method :)

custom_components/tahoma/climate_st.py
tetienne: That's the "ternary" operator I don't like when on multiple line. A classic if block is more readable.
```python
@property
    def preset_mode(self) -> Optional[str]:
        """Return the current preset mode, e.g., home, away, temp."""
       state = self.select_state(ST_DEROGATION_HEATING_MODE_STATE)
       if self.hvac_mode == HVAC_MODE_AUTO
         state =  self.select_state(ST_HEATING_MODE_STATE)
       return MAP_PRESET_MODES[st_preset]
```

custom_components/tahoma/climate_st.py
tetienne: Got it. HA probably has a check on this property if we set a value below this value.

custom_components/tahoma/climate_st.py
vlebourl: it attaches the temperature sensor to the thermostat entity. It's taken from the generic thermostat where you have to specify the temperature sensor: https://github.com/home-assistant/core/blob/60009ec2f96001fd95713f0a835feac77513d156/homeassistant/components/generic_thermostat/climate.py#L180

custom_components/tahoma/climate_st.py
tetienne: Understood. This refresh command is not sent for the other widgets?

It would be nice to group our commands in one call like Somfy does.

custom_components/tahoma/climate_st.py
vlebourl: let's agree to disagree on that ;-)
@iMicknl, can you help us choose between the solutions?

custom_components/tahoma/climate_st.py
vlebourl: I think it's specific to the somfy smart thermostat.

custom_components/tahoma/climate_st.py
vlebourl: done

custom_components/tahoma/climate_deh.py
```diff
@@ -0,0 +1,78 @@
...
+
+    @property
+    def target_temperature(self):
+        """Return the temperature we try to reach."""
+        return self.select_state(CORE_LEVEL_STATE)
+
+    def set_temperature(self, **kwargs) -> None:
+        """Set new target temperature."""
+        percentage = kwargs.get(ATTR_TEMPERATURE)
+        if percentage is None:
```
tetienne, line 55: It can happen?

vlebourl: not sure, took it from [here](https://github.com/home-assistant/core/blob/60009ec2f96001fd95713f0a835feac77513d156/homeassistant/components/generic_thermostat/climate.py#L330)

custom_components/tahoma/climate_deh.py
```diff
@@ -36,7 +40,7 @@ def supported_features(self) -> int:
     @property
     def temperature_unit(self) -> str:
         """Return the unit of measurement used by the platform."""
-        return TEMP_CELSIUS
+        return UNIT_PERCENTAGE
```
tetienne, line 43: So weird

vlebourl: I know... but the device is actually in percent, with 100 being of, and 0% being full on...
Thanks somfy for the logic

custom_components/tahoma/climate_deh.py
```diff
@@ -0,0 +1,84 @@
...
+        """Return the temperature we try to reach."""
+        return 100 - self.select_state(CORE_LEVEL_STATE)
+
+    def set_temperature(self, **kwargs) -> None:
+        """Set new target temperature."""
+        level = kwargs.get(ATTR_TEMPERATURE)
+        if level is None:
+            return
+        self.apply_action(COMMAND_SET_LEVEL, 100 - int(level))
+        self.apply_action(COMMAND_GET_LEVEL)
```
tetienne, line 62: If I understood correctly, that's what do Somfy on their website?

custom_components/tahoma/climate_deh.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def set_hvac_mode(self, hvac_mode: str) -> None:
+        """Set new target hvac mode."""
+        level = 0
+        if hvac_mode == HVAC_MODE_HEAT:
+            level = self._saved_level
+        else:
+            self._saved_level = self.target_temperature
+        self.apply_action(COMMAND_SET_LEVEL, 100 - int(level))
+        self.apply_action(COMMAND_GET_LEVEL)
```
tetienne, line 84: Sorry, I don't get the use case. Why do we need to store the target_temperature if not in HEAT mode?

custom_components/tahoma/climate_st.py
```diff
@@ -0,0 +1,245 @@
...
+    def set_temperature(self, **kwargs) -> None:
+        """Set new target temperature."""
+        temperature = kwargs.get(ATTR_TEMPERATURE)
+        if temperature is None:
+            return
+
+        if temperature < 15.0:
+            temperature = 15
+        elif temperature > 26.0:
+            temperature = 26
```
tetienne, line 202:
```suggestion
        if temperature < self.min_temp():
            temperature = self.min_temp()
        elif temperature > self.max_temp():
            temperature = self.max_temp()
```

custom_components/tahoma/climate_deh.py
vlebourl: So that we can restore the previous level when switching back on.

custom_components/tahoma/climate_deh.py
vlebourl: the state doesn't update on TaHoma, hence on HA. Pretty much like the smart thermostat and its `refreshState `command

PR #160, opened 2020-07-20
Make all functions async, utilise new Python wrapper, better exception handling, improved device mapping

iMicknl:
Depending on #159.


custom_components/tahoma/manifest.json
```diff
@@ -3,6 +3,8 @@
   "name": "Somfy TaHoma",
   "config_flow": true,
   "documentation": "https://www.home-assistant.io/integrations/tahoma",
-  "requirements": [],
+  "requirements": [
+    "git+https://github.com/iMicknl/python-tahoma-api.git#tahoma_api==0.1.0"
```
iMicknl, line 7: Eventually this should be changed to our PyPi package, but for the time being this will work and take our latest master.

custom_components/tahoma/__init__.py
```diff
@@ -36,52 +35,59 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
...
 
-    for component in PLATFORMS:
-        hass.async_create_task(
-            hass.config_entries.async_forward_entry_setup(entry, component)
-        )
-
-    for scene in scenes:
-        hass.data[DOMAIN][entry.entry_id]["scenes"].append(scene)
+    for platform in hass.data[DOMAIN][entry.entry_id]["entities"]:
+        if len(hass.data[DOMAIN][entry.entry_id]["entities"][platform]) > 0:
```
iMicknl, line 87: Now only platforms with existing entities will be set up. This makes our integration more efficient. I am not a fan of the if statement / loop structure, looking for suggestions.

custom_components/tahoma/tahoma_device.py
```diff
@@ -141,35 +147,33 @@ def has_command(self, *commands):
...
         return self.select_state(*states)
 
     @Throttle(timedelta(seconds=1))
-    def should_wait(self):
+    async def should_wait(self):
         """Wait for actions to finish."""
-        exec_queue = self.controller.get_current_executions()
-        self._exec_queue = [e for e in self._exec_queue if e in exec_queue]
+        exec_queue = await self.client.get_current_executions()
+        self._exec_queue = [e.id for e in self._exec_queue if e in exec_queue]
```
iMicknl, line 169: I am not 100% sure if this is functional. This should be tested more thoroughly.

custom_components/tahoma/tahoma_device.py
```diff
@@ -27,43 +29,45 @@
...
         """Return the name of the device."""
-        return self.tahoma_device.label
+        return self.device.label
 
     @property
     def available(self) -> bool:
         """Return True if entity is available."""
-        states = self.tahoma_device.active_states
+        # TODO Figure out how this works for RTS devices
+        if self.device.available is not None:
```
iMicknl, line 60: We didn't spot yet that the API also sends back `enabled` and `available` in the device information. We possibly can simplify this module, by removing the 3 if statements.

However I don't know if this `available` property covers all cases below and what it does for RTS devices.

custom_components/tahoma/tahoma_device.py
```diff
@@ -27,43 +29,45 @@
...
-    def update(self):
+    async def async_update(self):
         """Update method."""
-        if self.should_wait():
-            self.hass.add_job(self.update)
+        if await self.should_wait():
+            self.hass.async_add_job(self.async_update)
             return
-        self.controller.get_states([self.tahoma_device])
-        self.schedule_update_ha_state()
```
iMicknl, line 42: This looks not necessary? Do we know why this has been added here in the first place?

iMicknl:
@tetienne @vlebourl this is ready for a (draft) review. Please note that you should skip the climate files for now, due to #159 not merged yet. 

I think we should run this ourselves for a few days, to discover blocking bugs. Otherwise, we can merge this to master (or develop).

custom_components/tahoma/__init__.py
```diff
@@ -36,52 +35,59 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
...
+        _LOGGER.exception(exception)
+        return False
+    except Exception as exception:  # pylint: disable=broad-except
+        _LOGGER.exception(exception)
+        return False
+
+    devices = await client.get_devices()
+    scenes = await client.get_scenarios()
+
+    hass.data[DOMAIN][entry.entry_id] = {"controller": client, "entities": {}}
```
tetienne, line 57:
```suggestion
    hass.data[DOMAIN][entry.entry_id] = {"controller": client, "entities": {"scenes": scenes}}
```

custom_components/tahoma/__init__.py
```diff
@@ -36,52 +35,59 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
...
-    hass.data[DOMAIN][entry.entry_id] = {
-        "controller": controller,
-        "devices": [],
-        "scenes": [],
-    }
+    for platform in PLATFORMS:
+        hass.data[DOMAIN][entry.entry_id]["entities"][platform] = []
+
+    for scene in scenes:
+        hass.data[DOMAIN][entry.entry_id]["entities"]["scene"].append(scene)
```
tetienne, line 63: Useless.

custom_components/tahoma/__init__.py
tetienne: Look at default dict:

```python
>>> from collections import defaultdict
>>> d = defaultdict(list)
>>>d["foo"].append("faa")
``

File: custom_components/tahoma/__init__.py [hunk 8]
```diff
@@ -99,4 +105,16 @@ async def async_unload_entry(hass: HomeAssistant, entry: ConfigEntry):
...
     return unload_ok
+
+
+async def async_shutdown(self, _):
+    """Remove all future updates, cancel tasks and close the client."""
+    self._shutdown = True
+
+    for entry in self.hass.data[DOMAIN]:
+        client = entry.get("controller")
+        await client.close()
```
tetienne, line 120:
```suggestion
        await entry.get("controller").close()
```

custom_components/tahoma/config_flow.py
```diff
@@ -24,183 +17,37 @@
...
         errors = {}
 
         if user_input is not None:
-            unique_id = user_input.get(CONF_USERNAME)
-            await self.async_set_unique_id(unique_id)
+            username = user_input.get(CONF_USERNAME)
+            password = user_input.get(CONF_PASSWORD)
+
+            await self.async_set_unique_id(username)
             self._abort_if_unique_id_configured()
```
tetienne, line 35: Where it come from?

custom_components/tahoma/cover.py
```diff
@@ -101,16 +99,21 @@ class TahomaCover(TahomaDevice, CoverEntity):
...
 
         position = self.select_state(
             CORE_CLOSURE_STATE,
             CORE_DEPLOYMENT_STATE,
             CORE_PEDESTRIAN_POSITION_STATE,
             CORE_TARGET_CLOSURE_STATE,
         )
-        if position and "Horizontal" not in self.tahoma_device.widget:
+
+        if position is not None and "Horizontal" not in self.device.widget:
```
tetienne, line 114: Thx

custom_components/tahoma/cover.py
```diff
@@ -178,55 +181,62 @@ def device_state_attributes(self):
...
             else:
                 return ICON_LOCK_ALERT
+
         return None
 
-    def open_cover(self, **kwargs):
+    async def async_open_cover(self, **_):
         """Open the cover."""
-        self.apply_action(self.select_command(COMMAND_OPEN, COMMAND_UP))
+        await self.async_execute_command(self.select_command(COMMAND_OPEN, COMMAND_UP))
```
tetienne, line 209: Do we need this *async_execute_command*, what is its purpose?

custom_components/tahoma/manifest.json
tetienne: OKi. I think, our lib is ready for a first public version.

custom_components/tahoma/config_flow.py
iMicknl: @tetienne what do you mean? `user_input` was already there. Not is it just more structured with  both retrieval of `username` and `password` at the top of this function.

custom_components/tahoma/tahoma_device.py
tetienne: I hate this :D But @vlebourl like it :)

tetienne:
Awesome work! Did you notice any improvement in perf?

custom_components/tahoma/tahoma_device.py
iMicknl: @tetienne I can not read this myself as well... Thus not sure if it even works. It is still the old code.

iMicknl:
> Awesome work! Did you notice any improvement in perf?

Not really, since I don't have any devices nearby to test. Would be good if you and @vlebourl give it a try with all your devices for a while :).

custom_components/tahoma/__init__.py
iMicknl: Lol, good spot, this was in the original component as well... I have never spotted this as an useless loop. 👍

custom_components/tahoma/tahoma_device.py
tetienne: @vlebourl ?

vlebourl: which part? last line? 
`schedule_update_ha_state()` writes down the states to HA. maybe not necessary in an update method indeed.

iMicknl: I am not 100% sure, we should test this thoroughly.

custom_components/tahoma/__init__.py
iMicknl: @tetienne defaultdict is implemented! I still don't really like this big loop, any idea how to make this nicer?

tetienne: I think you have no choice here. You have to trigger the entry setup for all platform.

custom_components/tahoma/const.py
```diff
@@ -2,52 +2,61 @@
...
 SUPPORTED_PLATFORMS = [
-    "binary_sensor",
-    "climate",
-    "cover",
-    "light",
-    "lock",
-    "scene",
-    "sensor",
-    "switch",
+    BINARY_SENSOR,
```
tetienne, line 15: These contants already exist. See https://github.com/home-assistant/core/blob/dev/homeassistant/components/binary_sensor/__init__.py#L20 for instance.

custom_components/tahoma/tahoma_device.py
iMicknl: @vlebourl could you give this a try? Or pull the new API manually to understand the status of `available` for your RTS devices.

custom_components/tahoma/__init__.py
```diff
@@ -33,69 +22,89 @@ async def async_setup(hass: HomeAssistant, config: dict):
...
+        if len(entities_per_platform) > 0:
+            hass.async_create_task(
+                hass.config_entries.async_forward_entry_setup(entry, platform)
+            )
+
+    async def async_close_client(self, *_):
+        """Close HTTP client."""
+        await client.close()
+
+    hass.bus.async_listen_once(EVENT_HOMEASSISTANT_STOP, async_close_client)
```
tetienne, line 86: Nice!

custom_components/tahoma/config_flow.py
```diff
@@ -52,30 +28,29 @@ async def async_step_user(self, user_input=None):
...
 
+            client = TahomaClient(username, password)
```
tetienne, line 37: It will be nice to have a context manager for TahomaClient, it will avoid to think about the  close:
```Python
with TahomaClient(username, password) as client:
  client.login()
  ....
```

iMicknl: Will this auto close / take care of closing on exception? Or is that something that has to be built in the Python API :-).

iMicknl: @tetienne nevermind, I see your PR, looks good! Shall we do this improvement in a new PR later on? :)

iMicknl: Let's do this in a follow up PR!

tetienne: Yep

custom_components/tahoma/sensor.py
```diff
@@ -1,7 +1,7 @@
 """Support for TaHoma sensors."""
 from datetime import timedelta
 import logging
-from typing import Optional
+from typing import List, Optional
```
iMicknl, line 4: Why did you import list @vlebourl? Looks not used. :)

custom_components/tahoma/sensor.py
```diff
@@ -100,12 +100,17 @@ def state(self):
...
+            for attr in self.device.attributes
+            if attr.name == CORE_MEASURED_VALUE_TYPE
+        ]
         TEMPERATURE_UNIT = {
             CORE_TEMPERATURE_IN_CELSIUS: TEMP_CELSIUS,
             CORE_TEMPERATURE_IN_CELCIUS: TEMP_CELSIUS,
             CORE_TEMPERATURE_IN_KELVIN: TEMP_KELVIN,
             CORE_TEMPERATURE_IN_FAHRENHEIT: TEMP_FAHRENHEIT,
-        }.get(self.device.attributes.get(CORE_MEASURED_VALUE_TYPE), TEMP_CELSIUS,)
+        }.get(attribute[0], TEMP_CELSIUS,)
```
iMicknl, line 113: @vlebourl should we just use attribute.value here? That makes it more readable and nicely typed. Just use `attr` in the loop.

vlebourl: except `attribute` is a list, it doesn't have a `.value` attribute.

iMicknl: ` attribute = [
            attr.value
            for attr in self.device.attributes
            if attr.name == CORE_MEASURED_VALUE_TYPE
        ]
```

Attribute is not a list anymore? Otherwise we should change something in the loop, to just get the state back, so you can do attribute.value. Because taking the first element of a list doesn't sound safe to me?

tetienne: I think I will introduce a class States in the python-tahoma-api to add a getter method. We repeat a lot this pattern.

iMicknl: @tetienne agree, this is something we should solve in the wrapper. Shall we merge this in for now?

Issue #163, opened 2020-07-23
Sensors have too much precision.


vlebourl:
**Describe the bug**

Sensors report their values with full precision, making it hard to read

**Expected behavior**
limit to 2 digits precision.

**Screenshots**
![image](https://user-images.githubusercontent.com/5701372/88266780-140ece80-ccd0-11ea-869e-d93d11b3f0d9.png)
![image](https://user-images.githubusercontent.com/5701372/88266858-3acd0500-ccd0-11ea-8855-d3ffef3783dd.png)

**Environment (please complete the following information):**
 - **Home Assistant version:** 0.112.4
 - **Platform**: sensor
 - **ha-tahoma**: master (pre 1.7.0)

**Device:** (if your problem is related to a specific device)
Seen with io:TemperatureIOSystemSensor, somfythermostat:SomfyThermostatTemperatureSensor and somfythermostat:SomfyThermostatHumiditySensor

vlebourl:
closed with #164

PR #164, opened 2020-07-23
Round float state to 2 digits.

vlebourl:
Not sure whether we can get integer states. 
If not we can simplify by removing the if statement and rounding all the time.

Github Issue: #163


vlebourl:
Actually, python rounds an `int` without trouble. shall we just remove the if?

PR #168, opened 2020-07-24
Add more ui_classes and sort them alphabetically

iMicknl:
Merge into master after #160 is merged.


vlebourl:
isn't that a duplicate of #150?

iMicknl:
@vlebourl yes partly. However for now I just added the already supported platforms and I reordered the lists in the new structure.

vlebourl:
Nice. Do we really need to keep the other branch?

iMicknl:
@vlebourl I didn't want to close it for now, since you are also adding more classes as None. I am still not sure what I prefer / what would be nice. Let's have a look after the new integration is merged and discuss.

PR #177, opened 2020-07-27
Fix how the unit of measurement is retrieved

tetienne:
Taken from From https://www.tahomalink.com/enduser-mobile-web/steer-html5-client/tahoma/bootstrap.js

### DIFF
--- a/custom_components/tahoma/sensor.py
+++ b/custom_components/tahoma/sensor.py
@@ -1,21 +1,19 @@
 """Support for TaHoma sensors."""
 from datetime import timedelta
 import logging
-from typing import List, Optional
+from typing import Optional
 
 from homeassistant.components.sensor import DOMAIN as SENSOR
 from homeassistant.const import (
-    CONCENTRATION_PARTS_PER_MILLION,
     DEVICE_CLASS_HUMIDITY,
     DEVICE_CLASS_ILLUMINANCE,
     DEVICE_CLASS_POWER,
     DEVICE_CLASS_TEMPERATURE,
     ENERGY_WATT_HOUR,
-    POWER_WATT,
     TEMP_CELSIUS,
     TEMP_FAHRENHEIT,
     TEMP_KELVIN,
-    UNIT_PERCENTAGE,
+    VOLUME_CUBIC_METERS,
 )
 from homeassistant.helpers.entity import Entity
 
@@ -41,6 +39,7 @@ CORE_TEMPERATURE_IN_FAHRENHEIT = "core:TemperatureInFahrenheit"
 CORE_TEMPERATURE_IN_KELVIN = "core:TemperatureInKelvin"
 CORE_LUMINANCE_IN_LUX = "core:LuminanceInLux"
 CORE_ELECTRIC_ENERGY_IN_WH = "core:ElectricalEnergyInWh"
+CORE_VOLUME_IN_CUBIC_METER = "core:VolumeInCubicMeter"
 CORE_TEMPERATURE_STATE = "core:TemperatureState"
 CORE_THERMAL_ENERGY_CONSUMPTION_STATE = "core:ThermalEnergyConsumptionState"
 CORE_WINDSPEED_STATE = "core:WindSpeedState"
@@ -76,6 +75,7 @@ UNITS = {
     CORE_TEMPERATURE_IN_FAHRENHEIT: TEMP_FAHRENHEIT,
     CORE_LUMINANCE_IN_LUX: UNIT_LX,
     CORE_ELECTRIC_ENERGY_IN_WH: ENERGY_WATT_HOUR,
+    CORE_VOLUME_IN_CUBIC_METER: VOLUME_CUBIC_METERS,
 }
 
 
--- a/custom_components/tahoma/sensor.py
+++ b/custom_components/tahoma/sensor.py
@@ -110,13 +110,15 @@ class TahomaSensor(TahomaDevice, Entity):
             CORE_THERMAL_ENERGY_CONSUMPTION_STATE,
             CORE_WINDSPEED_STATE,
         )
-        return round(state, 2)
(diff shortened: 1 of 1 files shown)


custom_components/tahoma/sensor.py
```diff
@@ -104,35 +115,9 @@ def state(self):
...
-            CORE_TEMPERATURE_STATE: TEMPERATURE_UNIT,
-            CORE_RELATIVE_HUMIDITY_STATE: UNIT_PERCENTAGE,
-            CORE_LUMINANCE_STATE: UNIT_LX,
-            CORE_ELECTRIC_POWER_CONSUMPTION_STATE: POWER_WATT,
-            CORE_ELECTRIC_ENERGY_CONSUMPTION_STATE: ENERGY_WATT_HOUR,
-            CORE_CO_CONCENTRATION_STATE: CONCENTRATION_PARTS_PER_MILLION,
-            CORE_CO2_CONCENTRATION_STATE: CONCENTRATION_PARTS_PER_MILLION,
-        }.get(state)
+        attribute = self.device.attributes[CORE_MEASURED_VALUE_TYPE]
+        return UNITS.get(attribute.value) if attribute else None
```
vlebourl, line 119: isn't `.get(attribute.value)` already returning None as a default?

tetienne: You are right, I will simply this.

custom_components/tahoma/sensor.py
```diff
@@ -104,35 +115,8 @@ def state(self):
...
-            CORE_RELATIVE_HUMIDITY_STATE,
-            CORE_TEMPERATURE_STATE,
-        )
-        return {
-            CORE_TEMPERATURE_STATE: TEMPERATURE_UNIT,
-            CORE_RELATIVE_HUMIDITY_STATE: UNIT_PERCENTAGE,
-            CORE_LUMINANCE_STATE: UNIT_LX,
-            CORE_ELECTRIC_POWER_CONSUMPTION_STATE: POWER_WATT,
-            CORE_ELECTRIC_ENERGY_CONSUMPTION_STATE: ENERGY_WATT_HOUR,
-            CORE_CO_CONCENTRATION_STATE: CONCENTRATION_PARTS_PER_MILLION,
```
iMicknl, line 133: What about the units that can not be retrieved from the device attributes? It looks like this is removed in this PR?

custom_components/tahoma/sensor.py
vlebourl: my bad, won't work, `attribute.value` won't work if `attribute` is `None` cause `NoneType` has no attribute `value`

tetienne: Sorry, it was correct in fact. I have a sensor ('io:EnergyConsumptionSensorsHeatPumpComponent') which has no attributes field. I set this condition for such case.

custom_components/tahoma/sensor.py
tetienne: In such a case, we will return None. I don't introduce a regression. The previous code was not working. `select_state’ returns the value of a state. So in the case of my lux sensor, I got for instance 5356. So Celcius was always returned.
If you have other examples, I can easily add them.

iMicknl: In that case it has been broken since the last update.. We should make sure we move back to the original behaviour where we fallback to the available state to decide the unit.

https://github.com/iMicknl/ha-tahoma/blame/c6efc4c82bb01651827329659413eac23ff5ea10/custom_components/tahoma/sensor.py#L119

tetienne: You are right, the previous refactor changed the behavior. I will fix it. I found another measured unit here: https://github.com/iMicknl/ha-tahoma/issues/80
I think the way I retrieved the units make more sense. Somfy seems consistent here.

iMicknl: @tetienne I totally agree. Best would be to retrieve the value from the attributes. + a fallback to a hardcoded dict based on states that don't have those attributes.

tetienne:
Need to add m3. See https://github.com/redbug26/jeedom-tahoma/blob/6e53512c562ad980f83708313d602db1c05d844e/core/class/tahoma.class.php#L395

custom_components/tahoma/sensor.py
```diff
@@ -1,21 +1,19 @@
...
     DEVICE_CLASS_ILLUMINANCE,
     DEVICE_CLASS_POWER,
     DEVICE_CLASS_TEMPERATURE,
     ENERGY_WATT_HOUR,
-    POWER_WATT,
     TEMP_CELSIUS,
     TEMP_FAHRENHEIT,
     TEMP_KELVIN,
-    UNIT_PERCENTAGE,
+    VOLUME_CUBIC_METERS,
```
vlebourl, line 16: Why remove percentage? that's the `CORE_RELATIVE_HUMIDITY_STATE` unit.

tetienne: I was not able to find any device returning such MeasuredValueType. Even within Jeedom, there is no such case: https://github.com/redbug26/jeedom-tahoma/blob/6e53512c562ad980f83708313d602db1c05d844e/core/class/tahoma.class.php#L395

iMicknl: @tetienne we also miss `CONCENTRATION_PARTS_PER_MILLION` for CO and CO2 sensors.  And possibly also `POWER_WATT` or `ENERGY_KILO_WATT_HOUR`.

tetienne:
> CONCENTRATION_PARTS_PER_MILLION is missing, for the following sensors. Do they also have an attribute?
> 
> ```
> io:CO2IOSystemDeviceSensor
> io:CO2IOSystemSensor
> io:COIOSystemSensor
> ```

Sadly, I was not able to find any setup exemple with such exemple. I can ask on HA forum.

PR #180, opened 2020-07-27
Add device attributes to attributes

iMicknl:


iMicknl:
> Can you also remove this line please:
> 
> https://github.com/iMicknl/ha-tahoma/blob/426af899076bd3478cf3ebccbddb323636d32f93/custom_components/tahoma/tahoma_device.py#L82
> 
> 
> This last one is included with the next loop.

I don't think this is related to this PR, but I am also not sure if we should remove ATTR_RSSI_LEVEL which is a default named property from HA. The others are using the original names from HA, thus `core:RSSILevelState`.

PR #184, opened 2020-07-27
Fix cover implementation for devices with only cycle command

iMicknl:


tetienne:
It's weird but we have no choice.

iMicknl:
> It's weird but we have no choice.

Fully agree. Once we can utilise SUPPORT_TOGGLE as the only option, I happily move back!

PR #186, opened 2020-07-28
Implement the DataUpdateCoordinator

tetienne:
See #1

### DIFF
--- a/custom_components/tahoma/__init__.py
+++ b/custom_components/tahoma/__init__.py
@@ -44,7 +44,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
 
     scenes = await client.get_scenarios()
 
-    coordinator = TahomaDataUpdateCoordinator(
+    tahoma_coordinator = TahomaDataUpdateCoordinator(
         hass,
         _LOGGER,
         # Name of the data. For logging purposes.
@@ -55,17 +55,16 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
         update_interval=timedelta(seconds=5),
     )
 
-    await coordinator.async_refresh()
+    await tahoma_coordinator.async_refresh()
 
     hass.data[DOMAIN][entry.entry_id] = {
-        "controller": client,
         "entities": defaultdict(list),
-        "coordinator": coordinator,
+        "coordinator": tahoma_coordinator,
     }
 
     hass.data[DOMAIN][entry.entry_id]["entities"]["scene"] = scenes
 
-    for device in coordinator.data.values():
+    for device in tahoma_coordinator.data.values():
         if device.widget in TAHOMA_TYPES or device.ui_class in TAHOMA_TYPES:
             platform = TAHOMA_TYPES.get(device.widget) or TAHOMA_TYPES.get(
                 device.ui_class
--- a/custom_components/tahoma/__init__.py
+++ b/custom_components/tahoma/__init__.py
@@ -42,8 +42,6 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
         _LOGGER.exception(exception)
         return False
 
-    scenes = await client.get_scenarios()
-
     tahoma_coordinator = TahomaDataUpdateCoordinator(
         hass,
         _LOGGER,
@@ -52,7 +50,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
         client=client,
         devices=await client.get_devices(),
         listener_id=await client.register_event_listener(),
-        update_interval=timedelta(seconds=5),
+        update_interval=timedelta(seconds=30),
     )
 
     await tahoma_coordinator.async_refresh()
@@ -62,7 +60,9 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
         "coordinator": tahoma_coordinator,
     }
 
-    hass.data[DOMAIN][entry.entry_id]["entities"]["scene"] = scenes
+    hass.data[DOMAIN][entry.entry_id]["entities"][
+        "scene"
+    ] = await client.get_scenarios()
 
     for device in tahoma_coordinator.data.values():
         if device.widget in TAHOMA_TYPES or device.ui_class in TAHOMA_TYPES:
--- a/custom_components/tahoma/__init__.py
+++ b/custom_components/tahoma/__init__.py
@@ -4,7 +4,6 @@ from collections import defaultdict
 from datetime import timedelta
 import logging
 
-import async_timeout
 from tahoma_api.client import TahomaClient
 from tahoma_api.exceptions import BadCredentialsException, TooManyRequestsException
 
@@ -12,7 +11,6 @@ from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import CONF_PASSWORD, CONF_USERNAME, EVENT_HOMEASSISTANT_STOP
 from homeassistant.core import HomeAssistant
 
-from ...helpers.update_coordinator import DataUpdateCoordinator
 from .const import DOMAIN, SUPPORTED_PLATFORMS, TAHOMA_TYPES
 from .coordinator import TahomaDataUpdateCoordinator
 
--- a/custom_components/tahoma/__init__.py
+++ b/custom_components/tahoma/__init__.py
@@ -1,8 +1,10 @@
 """The TaHoma integration."""
 import asyncio
 from collections import defaultdict
+from datetime import timedelta
 import logging
 
+import async_timeout
 from tahoma_api.client import TahomaClient
 from tahoma_api.exceptions import BadCredentialsException, TooManyRequestsException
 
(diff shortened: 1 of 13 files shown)


custom_components/tahoma/TahomaDataUpdateCoordinator.py
```diff
@@ -0,0 +1,48 @@
...
+        self.devices: Dict[str, Device] = {d.deviceurl: d for d in devices}
+        self.listener_id = listener_id
+
+    async def _async_update_data(self) -> Dict[str, Device]:
+        """Fetch data from Tahoma."""
+        events = await self.client.fetch_event_listener(self.listener_id)
+        for event in events:
+            if event.name == "DeviceStateChangedEvent":
+                for state in event.device_states:
+                    self.devices[event.deviceurl].states[state.name].value = state.value
```
tetienne, line 47: TODO: cast according to the returned type.

tetienne:
I don't have fix the climate platform yet. @vlebourl I think I will need your help.

custom_components/tahoma/__init__.py
```diff
@@ -40,17 +44,30 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
...
+        client=client,
+        devices=await client.get_devices(),
+        listener_id=await client.register_event_listener(),
+        update_interval=timedelta(seconds=5),
+    )
+
+    await coordinator.async_refresh()
+
     hass.data[DOMAIN][entry.entry_id] = {
         "controller": client,
```
iMicknl, line 63: Could be removed? Looks not necessary anymore :).

custom_components/tahoma/__init__.py
```diff
@@ -40,17 +44,30 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
...
 
+    coordinator = TahomaDataUpdateCoordinator(
+        hass,
+        _LOGGER,
+        # Name of the data. For logging purposes.
+        name="Tahoma fetcher",
+        client=client,
+        devices=await client.get_devices(),
+        listener_id=await client.register_event_listener(),
+        update_interval=timedelta(seconds=5),
```
iMicknl, line 57: Is every 5 seconds not a bit to fast? Currently we only poll for updates when a device changes. This is great and will allow us to update the state if something happens outside HA.

But is every 30 seconds not enough? And maybe poll every second when we expect an update from a component, until we have received the new device state.

custom_components/tahoma/TahomaDataUpdateCoordinator.py
```diff
@@ -0,0 +1,48 @@
+from datetime import timedelta
```
iMicknl, line 1: Is the name of this file according to the python guidelines? Can we maybe just call it coordinator.py?

custom_components/tahoma/__init__.py
```diff
@@ -1,15 +1,19 @@
...
 
+import async_timeout
 from tahoma_api.client import TahomaClient
 from tahoma_api.exceptions import BadCredentialsException, TooManyRequestsException
 
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import CONF_PASSWORD, CONF_USERNAME, EVENT_HOMEASSISTANT_STOP
 from homeassistant.core import HomeAssistant
 
+from ...helpers.update_coordinator import DataUpdateCoordinator
```
iMicknl, line 15: Should be removed? It is breaking my build at least.

custom_components/tahoma/light.py
```diff
@@ -36,10 +36,11 @@ async def async_setup_entry(hass, entry, async_add_entities):
     """Set up the TaHoma lights from a config entry."""
```
iMicknl, line 36: Light has the same issue as cover.

```
return int(brightness * 255 / 100)
TypeError: unsupported operand type(s) for /: 'str' and 'int'
```

custom_components/tahoma/__init__.py
```diff
@@ -40,17 +44,30 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
...
         return False
 
-    devices = await client.get_devices()
     scenes = await client.get_scenarios()
 
+    coordinator = TahomaDataUpdateCoordinator(
+        hass,
+        _LOGGER,
+        # Name of the data. For logging purposes.
+        name="Tahoma fetcher",
```
iMicknl, line 53:
```suggestion
        name="TaHoma Event Fetcher",
```

custom_components/tahoma/climate_st.py
```diff
@@ -21,6 +23,7 @@
 from homeassistant.core import callback
 from homeassistant.helpers.event import async_track_state_change
 
+from . import TahomaDataUpdateCoordinator
```
iMicknl, line 26:
```suggestion
from .coordinator import TahomaDataUpdateCoordinator
```

custom_components/tahoma/__init__.py
tetienne: Yes 5 seconds was for testing, I will set it to 30 seconds. About the second remark, we can see this in a second PR. This one is already big enough.

custom_components/tahoma/__init__.py
tetienne: You are right.

custom_components/tahoma/light.py
tetienne: Fixed.

custom_components/tahoma/climate_st.py
tetienne: Done.

custom_components/tahoma/binary_sensor.py
```diff
@@ -61,10 +60,10 @@
 async def async_setup_entry(hass, entry, async_add_entities):
     """Set up the TaHoma sensors from a config entry."""
     data = hass.data[DOMAIN][entry.entry_id]
-    controller = data.get("controller")
+    coordinator = data.get("coordinator")
 
     entities = [
-        TahomaBinarySensor(device, controller)
+        TahomaBinarySensor(device.url, coordinator)
```
iMicknl, line 66: device.url?

custom_components/tahoma/tahoma_device.py
```diff
@@ -1,14 +1,12 @@
...
 
-from tahoma_api.client import TahomaClient
-from tahoma_api.models import Command, Device
+from tahoma_api.models import Command
 
 from homeassistant.const import ATTR_BATTERY_LEVEL
 from homeassistant.helpers.entity import Entity
-from homeassistant.util import Throttle
 
+from . import TahomaDataUpdateCoordinator
```
iMicknl, line 9:
```suggestion
from .coordinator import TahomaDataUpdateCoordinator
```

tetienne: That's isort that do it. I cannot change it.

custom_components/tahoma/coordinator.py
```diff
@@ -0,0 +1,60 @@
...
+            hass, logger, name=name, update_interval=update_interval,
+        )
+
+        self.client = client
+        self.devices: Dict[str, Device] = {d.deviceurl: d for d in devices}
+        self.listener_id = listener_id
+
+    async def _async_update_data(self) -> Dict[str, Device]:
+        """Fetch data from Tahoma."""
+        events = await self.client.fetch_event_listener(self.listener_id)
```
iMicknl, line 46: Should there be no try catch here? I see all integrations and samples use `            raise UpdateFailed()`. We can also have exceptions here which we should handle.

iMicknl:
https://github.com/iMicknl/ha-tahoma/blob/1aa76f03fa17ba49e319418f809ba33a0c2cd43d/custom_components/tahoma/__init__.py#L108

Here we should make use of the right client, this one is not existing anymore.

custom_components/tahoma/__init__.py
iMicknl: @tetienne @vlebourl maybe we should keep 5 seconds for now? Until we switch to an approach where we update once every x minutes and every second after an update. 

If we set it to 30 seconds, updates will still be slow.

custom_components/tahoma/binary_sensor.py
tetienne: You are right. I think I will create a PR request to rename device.deviceurl to device.url, make more sense and we don't repeat ourself.

custom_components/tahoma/coordinator.py
tetienne: Done.

tetienne:
> https://github.com/iMicknl/ha-tahoma/blob/1aa76f03fa17ba49e319418f809ba33a0c2cd43d/custom_components/tahoma/__init__.py#L108
> 
> Here we should make use of the right client, this one is not existing anymore.

Good catch.

custom_components/tahoma/__init__.py
tetienne: I'm afraid 5 seconds will overlead the HA, it's recommended to keep high. I was not able to find a component where such low poll interval.

About your suggestion to reduce it, when we just do an update using HA, we can see if we can schedule an update 5s later.

PR #196, opened 2020-07-30
Fix missing client for scene

tetienne:


tetienne:
> If you have tested this change and if it still works for all platforms, feel free to merge.

Tested on my  own scenes.

PR #197, opened 2020-07-30
Improve climate platform and support dynamic state allocation

tetienne:
@vlebourl Can you please validate the thermostat still work on your instance?

### DIFF
--- a/custom_components/tahoma/climate.py
+++ b/custom_components/tahoma/climate.py
@@ -7,11 +7,11 @@ from .climate_deh import DimmerExteriorHeating
 from .climate_st import SomfyThermostat
 from .const import DOMAIN
 
-AEH = "AtlanticElectricalHeater"
-ST = "SomfyThermostat"
-DEH = "DimmerExteriorHeating"
-
-TYPE = {AEH: AtlanticElectricalHeater, ST: SomfyThermostat, DEH: DimmerExteriorHeating}
+TYPE = {
+    "AtlanticElectricalHeater": AtlanticElectricalHeater,
+    "SomfyThermostat": SomfyThermostat,
+    "DimmerExteriorHeating": DimmerExteriorHeating,
+}
 
 
 async def async_setup_entry(hass, entry, async_add_entities):
--- a/custom_components/tahoma/climate.py
+++ b/custom_components/tahoma/climate.py
@@ -11,6 +11,8 @@ AEH = "AtlanticElectricalHeater"
 ST = "SomfyThermostat"
 DEH = "DimmerExteriorHeating"
 
+TYPE = {AEH: AtlanticElectricalHeater, ST: SomfyThermostat, DEH: DimmerExteriorHeating}
+
 
 async def async_setup_entry(hass, entry, async_add_entities):
     """Set up the TaHoma climate from a config entry."""
@@ -20,19 +22,9 @@ async def async_setup_entry(hass, entry, async_add_entities):
 
     climate_devices = [device for device in data.get("entities").get(CLIMATE)]
 
-    entities = []
-    for device in climate_devices:
-        if device.widget == AEH:
-            entities.append(AtlanticElectricalHeater(device.deviceurl, coordinator))
-        elif device.widget == ST:
-            base_url = device.deviceurl.split("#", 1)[0]
-            sensor = None
-            entity_registry = await hass.helpers.entity_registry.async_get_registry()
-            for k, v in entity_registry.entities.items():
-                if v.unique_id == base_url + "#2":
-                    sensor = k
-                    break
-            entities.append(SomfyThermostat(device.deviceurl, coordinator, sensor))
-        elif device.widget == DEH:
-            entities.append(DimmerExteriorHeating(device.deviceurl, coordinator))
+    entities = [
+        TYPE[device.widget](device.deviceurl, coordinator)
+        for device in climate_devices
+        if device.widget in TYPE
+    ]
     async_add_entities(entities)
--- a/custom_components/tahoma/climate_st.py
+++ b/custom_components/tahoma/climate_st.py
@@ -115,14 +115,20 @@ class SomfyThermostat(TahomaDevice, ClimateEntity):
             None,
         )
 
-        async_track_state_change(
-            self.hass, self._temp_sensor_entity_id, self._async_temp_sensor_changed
-        )
+        if self._temp_sensor_entity_id:
+            async_track_state_change(
+                self.hass, self._temp_sensor_entity_id, self._async_temp_sensor_changed
+            )
+
+        else:
+            _LOGGER.warning(
+                "Temperature sensor could not be found for entity %s", self.name
+            )
 
         @callback
         def _async_startup(event):
             """Init on startup."""
-            if self._temp_sensor_entity_id is not None:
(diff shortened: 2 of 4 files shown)


tetienne:
@vlebourl So did you test this branch?

vlebourl:
See various fixes in #203.
Requires: https://github.com/iMicknl/python-tahoma-api/pull/29

custom_components/tahoma/climate.py
```diff
@@ -11,6 +11,8 @@
 ST = "SomfyThermostat"
 DEH = "DimmerExteriorHeating"
 
+TYPE = {AEH: AtlanticElectricalHeater, ST: SomfyThermostat, DEH: DimmerExteriorHeating}
```
iMicknl, line 14: Any specific reason why this is not noted as:

```python
TYPE = {
	"AtlanticElectricalHeater": AtlanticElectricalHeater, 
	"SomfyThermostat": SomfyThermostat, 
	"DimmerExteriorHeating": DimmerExteriorHeating
}
```

Makes it a bit more readable for me, and easier to extend.

custom_components/tahoma/climate_st.py
```diff
@@ -99,6 +102,15 @@ def __init__(
     async def async_added_to_hass(self):
         """Register temperature sensor after added to hass."""
         await super().async_added_to_hass()
+
+        # The Somfy Thermostat require a temperature sensor
+        base_url = self.device.deviceurl.split("#", 1)[0]
+        entity_registry = await self.hass.helpers.entity_registry.async_get_registry()
+        for k, v in entity_registry.entities.items():
```
iMicknl, line 109: what is `k` and `v`? (besides key/value). Would be good to add a bit more meaningful names to improve readability. Since it is not a plain key value list, but it is reflecting a device?

custom_components/tahoma/climate_st.py
```diff
@@ -99,6 +102,15 @@ def __init__(
...
+            if v.unique_id == f"{base_url}#2":
+                self._temp_sensor_entity_id = k
```
iMicknl, line 111: Can we not just say
```python
next(sensor for sensor in entity_registry.entities.items() if sensor.unique_id == f"{base_url}#2")
```

custom_components/tahoma/climate_st.py
```diff
@@ -99,6 +102,15 @@ def __init__(
     async def async_added_to_hass(self):
         """Register temperature sensor after added to hass."""
         await super().async_added_to_hass()
+
+        # The Somfy Thermostat require a temperature sensor
```
iMicknl, line 106:
```suggestion
        # The Somfy Thermostat requires a temperature sensor
```

custom_components/tahoma/climate.py
vlebourl: because the words were long to type, but fair enough

custom_components/tahoma/climate_st.py
vlebourl: for loop removed following next comment.

custom_components/tahoma/climate_st.py
vlebourl: done (with a fix)

custom_components/tahoma/climate_st.py
```diff
@@ -99,6 +102,16 @@ def __init__(
     async def async_added_to_hass(self):
         """Register temperature sensor after added to hass."""
         await super().async_added_to_hass()
+
+        # The Somfy Thermostat requires a temperature sensor
+        base_url = self.device.deviceurl.split("#", 1)[0]
+        entity_registry = await self.hass.helpers.entity_registry.async_get_registry()
+        self._temp_sensor_entity_id = next(
+            entity_id
```
iMicknl, line 110: Nice change!! I just wanted to respond that the other notation was not really clear haha. :-)

What is the result when next() can not find a sensor? Will it be None? Otherwise, do we need to add a `None` fallback. https://stackoverflow.com/questions/8653516/python-list-of-dictionaries-search

vlebourl: "If the default parameter is omitted and the iterator is exhausted, it raises StopIteration exception."

I guess `None` is ok as a fallback.

tetienne: If you set None as default value you will have to protect the code each time the sensor is used. Normally it should not happen but I feel the exception will be more useful.

vlebourl: you're right, as is, with None as a default, next line will fail with:
```python

2020-08-05 12:22:06 ERROR (MainThread) [homeassistant.components.climate] Error while setting up tahoma platform for climate
  File "/Users/vlb/PycharmProjects/ha-tahoma/venv/lib/python3.8/site-packages/homeassistant/helpers/entity_platform.py", line 187, in _async_setup_platform
    await asyncio.gather(*pending)
  File "/Users/vlb/PycharmProjects/ha-tahoma/venv/lib/python3.8/site-packages/homeassistant/helpers/entity_platform.py", line 293, in async_add_entities
    await asyncio.gather(*tasks)
  File "/Users/vlb/PycharmProjects/ha-tahoma/venv/lib/python3.8/site-packages/homeassistant/helpers/entity_platform.py", line 450, in _async_add_entity
    await entity.async_added_to_hass()
  File "/Users/vlb/PycharmProjects/ha-tahoma/Config/custom_components/tahoma/climate_st.py", line 118, in async_added_to_hass
    async_track_state_change(
  File "/Users/vlb/PycharmProjects/ha-tahoma/venv/lib/python3.8/site-packages/homeassistant/helpers/event.py", line 92, in async_track_state_change
    entity_ids = tuple(entity_id.lower() for entity_id in entity_ids)
TypeError: 'NoneType' object is not iterable
```

One solution is to only track change line 118 if sensor isn't None, or to fail wit the StopIteration exception.

vlebourl: second solution is better to me, it allows the component to start gracefully and shows the climate device in the UI, just without the current temperature.
image

iMicknl: Would it make sense to raise a custom warning for the logging, which makes it easier to debug?

vlebourl: sure, that's ok with me

iMicknl: I don't know anything about climate devices... But I would opt for a way where we start gracefully indeed and just skip the temperature. Maybe add a warning to the log that it could not find the temperature device.

iMicknl: However, this device always has a temperature sensor I believe? Thus we are looking at a super edge case at the moment. Not sure if we need to spend a lot of work on it.

vlebourl: Indeed, it's part of the physical device, so unless they change the url of the device, I don't see how it could not be found...

vlebourl: this would start gracefully in such a case:
```py
if self._temp_sensor_entity_id:
    async_track_state_change(
        self.hass, self._temp_sensor_entity_id, self._async_temp_sensor_changed
    )
else:
    _LOGGER.warning("Temperature sensor could not be found for entity %s", self.name)
```

vlebourl:
It's been tested, let's wait for @tetienne approval.

PR #198, opened 2020-07-30
Reduce platform setup complexity

### LINKED ISSUE #191: Error fetching TaHoma Event Fetcher data: Error communicating with the TaHoma API: None
**Describe the bug**
Error fetching TaHoma Event Fetcher data: Error communicating with the TaHoma API: None

**To Reproduce**

1. Start HomeAssistant

**Expected behavior**
No error :)

**Environment (please complete the following information):**

- **Home Assistant version:** 0.113.2

**Device:** (if your problem is related to a specific device)

Using version 1ef0944

Just reporting this in case it's useful

tetienne:
I've also increased the poll interval to 10.

Fix #191

### DIFF
--- a/custom_components/tahoma/__init__.py
+++ b/custom_components/tahoma/__init__.py
@@ -7,15 +7,16 @@ import logging
 from pyhoma.client import TahomaClient
 from pyhoma.exceptions import BadCredentialsException, TooManyRequestsException
 
+from homeassistant.components.scene import DOMAIN as SCENE
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import CONF_PASSWORD, CONF_USERNAME, EVENT_HOMEASSISTANT_STOP
 from homeassistant.core import HomeAssistant
 
-from .const import DOMAIN, SUPPORTED_PLATFORMS, TAHOMA_TYPES
+from .const import DOMAIN, TAHOMA_TYPES
 from .coordinator import TahomaDataUpdateCoordinator
 
 _LOGGER = logging.getLogger(__name__)
-DEFAULT_UPDATE_INTERVAL = timedelta(seconds=5)
+DEFAULT_UPDATE_INTERVAL = timedelta(seconds=10)
 
 
 async def async_setup(hass: HomeAssistant, config: dict):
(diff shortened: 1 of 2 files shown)


custom_components/tahoma/__init__.py
```diff
@@ -7,15 +7,16 @@
 from pyhoma.client import TahomaClient
 from pyhoma.exceptions import BadCredentialsException, TooManyRequestsException
 
+from homeassistant.components.scene import DOMAIN as SCENE
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import CONF_PASSWORD, CONF_USERNAME, EVENT_HOMEASSISTANT_STOP
 from homeassistant.core import HomeAssistant
 
-from .const import DOMAIN, SUPPORTED_PLATFORMS, TAHOMA_TYPES
+from .const import DOMAIN, TAHOMA_TYPES
```
iMicknl, line 15: Why is `SUPPORTED_PLATFORMS` removed? Would be good to keep it, instead of just relying on the platforms defined in const.py.

tetienne: The supported platforms are including in TAHOMA_TYPES, and obviously we support all of them. So this [condition](https://github.com/iMicknl/ha-tahoma/pull/198/files#diff-d28fdb4b71ecdf95d6ca1c2a918de26bL73) does not make sense, never we will map a device we don't support.

iMicknl: I like the simplified code a lot more 👍. My only concern was that if we move to core, we need to carefully check if there are no extra mappings for integrations not yet present in those PRs.

tetienne: >we need to carefully check if there are no extra mappings for integrations not yet present in those PRs. 

I don't get it.

iMicknl: @tetienne a PR to core will not include all platforms, thus we need to be careful with the list containing the mapping.

tetienne: Oh I see. My goal is to clean as much as possible our code to make it more readable and much easier to review.

PR #200, opened 2020-07-31
Create constant for ignored devices

tetienne:


custom_components/tahoma/const.py
```diff
@@ -9,6 +9,12 @@
 
 DOMAIN = "tahoma"
 
+TAHOMA_NAMES_IGNORED = [
+    "ogp:Bridge",
+    "internal:PodV2Component",
+    "internal:TSKAlarmComponent",
```
tetienne, line 15: To check if we can not use the widget or ui class.

iMicknl: That was exactly my thought as well! Only for alarm we would need to scope a bit better, but perhaps there is a widget for internal alarms.

iMicknl:
```
"qualifiedName": "internal:TSKAlarmComponent",
"widgetName": "TSKAlarmController",
"uiClass": "Alarm",
```

```
"qualifiedName": "internal:PodV2Component",
"widgetName": "Pod",
"uiClass": "Pod",
```

```
"qualifiedName": "ogp:Bridge",
"widgetName": "DynamicBridge",
"uiClass": "ProtocolGateway",
```

Best would be to match on `uiClass` and/or `widgetName`. For the Alarm, we would need to match on widgetName is TSKAlarmController.

iMicknl: @tetienne I took the liberty to make a small change to your branch myself. :-)

PR #203, opened 2020-08-03
Some states are missing at init in tahoma's response

vlebourl:
Missing states must be dynamically added later.

requires: https://github.com/iMicknl/python-tahoma-api/pull/29

### DIFF
--- a/custom_components/tahoma/climate_st.py
+++ b/custom_components/tahoma/climate_st.py
@@ -4,6 +4,8 @@ from typing import List, Optional
 
 from homeassistant.components.climate import ClimateEntity
 from homeassistant.components.climate.const import (
+    CURRENT_HVAC_HEAT,
+    CURRENT_HVAC_IDLE,
     HVAC_MODE_AUTO,
     HVAC_MODE_HEAT,
     PRESET_AWAY,
@@ -85,9 +87,12 @@ class SomfyThermostat(TahomaDevice, ClimateEntity):
         super().__init__(device_url, coordinator)
         self._temp_sensor_entity_id = None
         if self.hvac_mode == HVAC_MODE_AUTO:
-            self._saved_target_temp = self.select_state(
-                PRESET_TEMPERATURES[self.preset_mode]
-            )
+            if self.preset_mode == PRESET_NONE:
+                self._saved_target_temp = None
+            else:
+                self._saved_target_temp = self.select_state(
+                    PRESET_TEMPERATURES[self.preset_mode]
+                )
         else:
             self._saved_target_temp = self.select_state(
                 CORE_DEROGATED_TARGET_TEMPERATURE_STATE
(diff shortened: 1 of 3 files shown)


custom_components/tahoma/coordinator.py
```diff
@@ -52,12 +53,15 @@ async def _async_update_data(self) -> Dict[str, Device]:
                 if event.name == "DeviceStateChangedEvent":
                     for state in event.device_states:
                         device = self.devices[event.deviceurl]
-                        device.states[state.name].value = self._get_state(state)
+                        if device.states[state.name] is None:
```
tetienne, line 56:
```suggestion
                        if state.name not in device.states:
```

custom_components/tahoma/coordinator.py
```diff
@@ -52,12 +53,15 @@ async def _async_update_data(self) -> Dict[str, Device]:
                 if event.name == "DeviceStateChangedEvent":
                     for state in event.device_states:
                         device = self.devices[event.deviceurl]
-                        device.states[state.name].value = self._get_state(state)
+                        if state.name not in device.states:
```
tetienne, line 56:
```suggestion
                        if state.name not in device.states:
```

custom_components/tahoma/coordinator.py
```diff
@@ -52,12 +53,15 @@ async def _async_update_data(self) -> Dict[str, Device]:
...
+                            device.states[state.name] = state
+                        else:
```
tetienne, line 58: You can remove the else. This way with the next line we will cast the state value.

custom_components/tahoma/climate_st.py
```diff
@@ -85,9 +87,12 @@ def __init__(self, device_url: str, coordinator: TahomaDataUpdateCoordinator):
...
         if self.hvac_mode == HVAC_MODE_AUTO:
-            self._saved_target_temp = self.select_state(
-                PRESET_TEMPERATURES[self.preset_mode]
-            )
+            if self.preset_mode == PRESET_NONE:
+                self._saved_target_temp = None
+            else:
+                self._saved_target_temp = self.select_state(
+                    PRESET_TEMPERATURES[self.preset_mode]
+                )
```
tetienne, line 95: Can you please what it means graphically?

custom_components/tahoma/climate_st.py
```diff
@@ -160,6 +165,15 @@ def hvac_modes(self) -> List[str]:
         """Return the list of available hvac operation modes."""
         return [HVAC_MODE_AUTO, HVAC_MODE_HEAT]
 
+    @property
+    def hvac_action(self) -> str:
```
tetienne, line 169: Cool mode is not supported?

vlebourl: no. this is a heater, not an AC

custom_components/tahoma/climate_st.py
vlebourl: If target is None, the target temp just doesn't show on the card, and the slider stays at the bottom left end.

iMicknl:
@vlebourl could you 'resolve conversations' when you refactored based on review? (and leave them open in case there is an ongoing discussion). Makes it easier to review, after a previous review :-)

I am ok by merging this one to @tetienne's branch, where we can review again and merge to master.

PR #213, opened 2020-08-11
Add first unit tests for Config Flow

iMicknl:


iMicknl:
@vlebourl @tetienne sorry for the commit spam, had to test the GitHub Actions here... 95% test coverage for the config flow now! (I just don't know how to fix the remaining 5%)

Screenshot 2020-08-11 at 11 17 13

tests/test_config_flow.py
```diff
@@ -0,0 +1,163 @@
...
+    
+    result = await hass.config_entries.flow.async_init(
+        config_flow.DOMAIN, context={"source": config_entries.SOURCE_USER}
+    )
+
+    assert result["type"] == "form"
+    assert result["errors"] == {}
+
+    with patch("pyhoma.client.TahomaClient.login", return_value=True), patch(
+        "custom_components.tahoma.async_setup", return_value=True
```
tetienne, line 23: We have to mock this because we don't have a true hass instance?

tests/test_config_flow.py
```diff
@@ -0,0 +1,163 @@
...
+    ) as mock_setup, patch(
+        "custom_components.tahoma.async_setup_entry", return_value=True
+    ) as mock_setup_entry:
+        result2 = await hass.config_entries.flow.async_configure(
+            result["flow_id"],
+            {"username": "test@testdomain.com", "password": "test-password"},
+        )
+
+    assert result2["type"] == "create_entry"
```
tetienne, line 32: You can do one assert:
```Python
assert result2 == {"type": "create_entry", "title": "test@testdomain.com", ...}
```
except if their is some random value in result2.

tests/test_config_flow.py
```diff
@@ -0,0 +1,163 @@
...
+        assert result["title"] == "test@testdomain.nl"
+        assert result["data"] == {
+            "username": "test@testdomain.nl",
+            "password": "test-password",
+        }
+
+        await hass.async_block_till_done()
+
+        assert len(mock_setup.mock_calls) == 1
+        assert len(mock_setup_entry.mock_calls) == 1
```
tetienne, line 163: Create a constant EMAIL_1 set to test@testdomain.com and EMAIL_2 to avoid to repeat them, it will ease the test also.

tests/test_config_flow.py
```diff
@@ -0,0 +1,163 @@
...
+    )
+
+    with patch("pyhoma.client.TahomaClient.login", side_effect=Exception):
+        result2 = await hass.config_entries.flow.async_configure(
+            result["flow_id"],
+            {"username": "test@testdomain.com", "password": "test-password"},
+        )
+
+    assert result2["type"] == "form"
+    assert result2["errors"] == {"base": "unknown"}
```
tetienne, line 89: They are basically the same tests. Only the message and exception change. Have a look to parametrize test with Pytest.

tetienne:
> @vlebourl @tetienne sorry for the commit spam, had to test the GitHub Actions here... 95% test coverage for the config flow now! (I just don't know how to fix the remaining 5%)
> 
> Screenshot 2020-08-11 at 11 17 13

If you can make `client.login()` raise an error it should be ok.

tests/test_config_flow.py
iMicknl: Yes, this was part of the test boilerplate in core when running hassfest. And it is also how most components (that I checked) handle this.

tests/test_config_flow.py
iMicknl: I prefer to match them on a per key base to easily understand which part is failing and I also believe that we can not change the order in your example, without breaking the test?

Does this make sense?

pytest.ini
```diff
@@ -0,0 +1,4 @@
+[pytest]
+filterwarnings =
+    # respx/asynctest: https://github.com/lundberg/respx/issues/61
+    ignore:"@coroutine" decorator is deprecated since Python 3.8, use "async def" instead:DeprecationWarning
```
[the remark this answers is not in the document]
iMicknl: I need to have a better look, but on my local machine I needed them. It looks like somewhere in Home Assistant or pytest-homeassistant they are using @coroutine.

tests/test_config_flow.py
tetienne: Pytest have a nice output with the json assertion. It does not care about the order (except if an array). Have a try ;)

iMicknl:
```
E       AssertionError: assert {'data': {'pa...8c074b1', ...} == {'data': {'pa...create_entry'}
E         Omitting 2 identical items, use -vv to show
E         Differing items:
E         {'data': {'password': 'test-password', 'username': 'test@testdomain.com'}} != {'data': {'password': '2test-password', 'username': 'test@testdomain.com'}}
E         Left contains 6 more items:
E         {'description': None,
E          'description_placeholders': None,
E          'flow_id': 'cb53d0783e9b40768027187808c074b1',...
E         
```

There are more items in the list apparently, which is why the assert doesn't work otherwise. This is also how they do it in the other components.

iMicknl:
Coverage is now 💯!! And by adding the latest test, I also spotted a bug and something that is not super nice. I have improved the error logging for YAML imports and I abort the flow correctly now.

custom_components/tahoma/config_flow.py
```diff
@@ -59,5 +59,12 @@ async def async_step_import(self, import_config: dict):
         if import_config:
             try:
                 return await self.async_validate_input(import_config)
+            except TooManyRequestsException:
+                _LOGGER.error("too_many_requests")
+                return self.async_abort(reason="too_many_requests")
+            except BadCredentialsException:
+                _LOGGER.error("invalid_auth")
+                return self.async_abort(reason="invalid_auth")
```
tetienne, line 67: You duplicate this code. You can move it within  _async_validate_input_

iMicknl: I wanted to just call `self.async_step_user(import_config)` from `async_step_import`, however there are some differences... The async_step_user works with a form and `errors` variable, where the import flow needs to write errors to the logging and `async_abort` the flow.

I couldn't find an elegant way to combine it and I expect the YAML flow to be deprecated in some versions, thus making it easy to remove.

tetienne:
That's a great work! We can be more confident now!

Issue #221, opened 2020-08-12
Add support for SomfyOneplusCamera (myfox:SomfyOnePlusCameraController)


PoVPoV29:
2020-08-12 15:14:24 DEBUG (MainThread) [custom_components.tahoma] Unsupported TaHoma device detected (myfox:SomfyOnePlusCameraController - Camera - MyFoxSecurityCamera)

**Device information**
The equipement is a Somy One plus Camera ("plus" is because there is a batery inside)
Can you integrate it with : 
- photo/image support
-on/off alarm (with status)
- activate siren
- detection status
- power status

**Device details**

```
{
	"commands": [{
		"commandName": "close",
		"nparams": 0
	}, {
		"commandName": "getName",
		"nparams": 0
	}, {
		"commandName": "open",
		"nparams": 0
	}, {
		"commandName": "refreshDeviceStatus",
		"nparams": 0
	}, {
		"commandName": "setName",
		"nparams": 1
	}, {
		"commandName": "refreshShutterStatus",
		"nparams": 0
	}],
	"states": [{
		"values": ["offline", "online", "uninstalled"],
		"type": "DiscreteState",
		"qualifiedName": "core:CloudDeviceStatusState"
	}, {
		"type": "DataState",
		"qualifiedName": "core:NameState"
	}, {
		"values": ["down", "up"],
		"type": "DiscreteState",
		"qualifiedName": "core:PowerSupplyUpDownState"
	}, {
		"values": ["closed", "opened"],
		"type": "DiscreteState",
		"qualifiedName": "myfox:ShutterStatusState"
	}],
	"dataProperties": [],
	"widgetName": "MyFoxSecurityCamera",
	"uiClass": "Camera",
	"qualifiedName": "myfox:SomfyOnePlusCameraController",
	"type": "ACTUATOR"
}
```

PoVPoV29:
Maybe it's releated with :
2020-08-12 15:14:24 DEBUG (MainThread) [custom_components.tahoma] Unsupported TaHoma device detected (myfox:SomfyProtectAlarmController - Alarm - MyFoxAlarmController)
2020-08-12 15:14:24 DEBUG (MainThread) [custom_components.tahoma] Unsupported TaHoma device detected (rtds:RTDSRemoteControllerComponent - RemoteController - AlarmRemoteController)

I don't know the meening of these two lines.

iMicknl:
[...]

PR #224, opened 2020-08-12
Support opening and closing the cover of MyFoxSecurityCamera 

iMicknl:
@tetienne you mapped it to a switch, however looking at the current codebase, it makes more sense to map it to `cover`. What do you think?

### DIFF
--- a/custom_components/tahoma/const.py
+++ b/custom_components/tahoma/const.py
@@ -40,6 +40,7 @@ TAHOMA_TYPES = {
     "Light": LIGHT,
     "LightSensor": SENSOR,
     "MotionSensor": BINARY_SENSOR,
+    "MyFoxSecurityCamera": COVER,  # widgetName, uiClass is Camera (not supported)
     "OccupancySensor": BINARY_SENSOR,
     "OnOff": SWITCH,
     "Pergola": COVER,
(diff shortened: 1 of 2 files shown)


tetienne:
Only open and close action are available. You cannot set the position. It's why I set it as a switch. It makes more sense IMO.

iMicknl:
> Only open and close action are available. You cannot set the position. It's why I set it as a switch. It makes more sense IMO.

Setting the position is not required for the `cover` platform. We also support other stateless Somfy devices in the cover platform. Imo it makes more sense to utilise the cover platform where all commands are supported already. (and the camera shutter is some kind of a cover)

We moved the stateless Garage Door also from the switch to the cover component for the reason above.

tetienne:
If I follow your logic the switch platform can be deleted as the cover platform can do the same.
Can you please add in comment an exemple of camera shutter setup?

iMicknl:
> If I follow your logic the switch platform can be deleted as the cover platform can do the same.
> Can you please add in comment an exemple of camera shutter setup?

Not really. My logic would be to map a cover (or device with cover like features and the same commands) to the cover platform, where the switch platform is used as an on/off switch. We map devices with only on / off commands to switch for example, and we modified it to also support a basic siren. (until there is a native siren platform).

For me, it doesn't make sense to implement `open`, `close` and `closed/opened` state in Switch as well. Especially since it is covered by the cover platform already, and the use case is a camera cover. https://github.com/iMicknl/ha-tahoma/issues/221#issue-677721254

We could add a new device class for a CAMERA_SHUTTER / CAMERA_COVER, however I am not sure if that is necessary.

tetienne:
OK looking at the setup it make sense. We already support these commands indeed.
Sorry for my comments.

iMicknl:
> OK looking at the setup it make sense. We already support these commands indeed.
> Sorry for my comments.

No worries, it is good to be challenged and to discuss architectural choices. 👍🏻. Thanks for your review!

PR #227, opened 2020-08-14
Improve entity update after command execution, add retry logic

iMicknl:
Currently, the default execution is set to 10 seconds, which is terribly slow when turning on/off a light. It would be good to poll every second after a command is executed, until the execution is finished.

This will allow us to poll the event listener every minute or hour for installations where the remote is not / rarely used. By making it configurable via an Option Flow, users adjust it to their scenario.

## ✨ Enhancement
* Make scan interval configurable
* Visually show if covers are opening or closing
* Improve entity update speed after command execution via HA

## 🐛 Bug Fixes
* Close client on error

Screenshot 2020-08-15 at 21 20 55
Screenshot 2020-08-20 at 14 33 24

### DIFF
--- a/custom_components/tahoma/__init__.py
+++ b/custom_components/tahoma/__init__.py
@@ -106,7 +106,6 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
         name="TaHoma Event Fetcher",
         client=client,
         devices=await client.get_devices(),
-        listener_id=await client.register_event_listener(),
         update_interval=timedelta(seconds=update_interval),
     )
 
--- a/custom_components/tahoma/__init__.py
+++ b/custom_components/tahoma/__init__.py
@@ -4,6 +4,7 @@ from collections import defaultdict
 from datetime import timedelta
 import logging
 
+from aiohttp import CookieJar
 from pyhoma.client import TahomaClient
 from pyhoma.exceptions import BadCredentialsException, TooManyRequestsException
 import voluptuous as vol
@@ -18,7 +19,7 @@ from homeassistant.const import (
     EVENT_HOMEASSISTANT_STOP,
 )
 from homeassistant.core import HomeAssistant
-from homeassistant.helpers import config_validation as cv
+from homeassistant.helpers import aiohttp_client, config_validation as cv
 
 from .config_flow import CONF_UPDATE_INTERVAL
 from .const import DEFAULT_UPDATE_INTERVAL, DOMAIN, IGNORED_TAHOMA_TYPES, TAHOMA_TYPES
@@ -76,7 +77,11 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
     username = entry.data.get(CONF_USERNAME)
     password = entry.data.get(CONF_PASSWORD)
 
-    client = TahomaClient(username, password)
+    session = aiohttp_client.async_create_clientsession(
+        hass, cookie_jar=CookieJar(unsafe=True)
+    )
+
+    client = TahomaClient(username, password, session=session)
 
     try:
         await client.login()
--- a/custom_components/tahoma/__init__.py
+++ b/custom_components/tahoma/__init__.py
@@ -87,15 +87,12 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
         await client.login()
     except TooManyRequestsException:
         _LOGGER.error("too_many_requests")
-        await client.close()
         return False
     except BadCredentialsException:
         _LOGGER.error("invalid_auth")
-        await client.close()
         return False
     except Exception as exception:  # pylint: disable=broad-except
         _LOGGER.exception(exception)
-        await client.close()
         return False
 
     update_interval = entry.options.get(CONF_UPDATE_INTERVAL, DEFAULT_UPDATE_INTERVAL)
@@ -140,19 +137,12 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
             hass.config_entries.async_forward_entry_setup(entry, platform)
         )
 
-    async def async_close_client(self, *_):
-        """Close HTTP client."""
-        await client.close()
-
-    hass.bus.async_listen_once(EVENT_HOMEASSISTANT_STOP, async_close_client)
-
     return True
 
 
 async def async_unload_entry(hass: HomeAssistant, entry: ConfigEntry):
     """Unload a config entry."""
 
-    await hass.data[DOMAIN][entry.entry_id].get("coordinator").client.close()
     await hass.data[DOMAIN][entry.entry_id].get("update_listener")()
 
     entities_per_platform = hass.data[DOMAIN][entry.entry_id]["entities"]
--- a/custom_components/tahoma/__init__.py
+++ b/custom_components/tahoma/__init__.py
@@ -143,8 +143,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
 async def async_unload_entry(hass: HomeAssistant, entry: ConfigEntry):
     """Unload a config entry."""
 
-    await hass.data[DOMAIN][entry.entry_id].get("update_listener")()
-
+    hass.data[DOMAIN][entry.entry_id]["update_listener"]()
     entities_per_platform = hass.data[DOMAIN][entry.entry_id]["entities"]
 
     unload_ok = all(
--- a/custom_components/tahoma/__init__.py
+++ b/custom_components/tahoma/__init__.py
@@ -56,7 +56,7 @@ async def async_setup(hass: HomeAssistant, config: dict):
         configuration.get(CONF_USERNAME) in entry.data.get(CONF_USERNAME)
         for entry in hass.config_entries.async_entries(DOMAIN)
     ):
-        return True
+        return False
 
     hass.async_create_task(
         hass.config_entries.flow.async_init(
--- a/custom_components/tahoma/__init__.py
+++ b/custom_components/tahoma/__init__.py
@@ -16,8 +16,13 @@ from homeassistant.const import CONF_EXCLUDE, CONF_PASSWORD, CONF_USERNAME
 from homeassistant.core import HomeAssistant
 from homeassistant.helpers import aiohttp_client, config_validation as cv
 
-from .config_flow import CONF_UPDATE_INTERVAL
-from .const import DEFAULT_UPDATE_INTERVAL, DOMAIN, IGNORED_TAHOMA_TYPES, TAHOMA_TYPES
+from .const import (
+    CONF_UPDATE_INTERVAL,
+    DEFAULT_UPDATE_INTERVAL,
+    DOMAIN,
+    IGNORED_TAHOMA_TYPES,
+    TAHOMA_TYPES,
+)
 from .coordinator import TahomaDataUpdateCoordinator
 
 _LOGGER = logging.getLogger(__name__)
--- a/custom_components/tahoma/__init__.py
+++ b/custom_components/tahoma/__init__.py
@@ -10,7 +10,6 @@ import voluptuous as vol
 
 from homeassistant import config_entries
 from homeassistant.components.scene import DOMAIN as SCENE
-from homeassistant.components.tahoma.config_flow import CONF_UPDATE_INTERVAL
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import (
     CONF_EXCLUDE,
@@ -21,6 +20,7 @@ from homeassistant.const import (
 from homeassistant.core import HomeAssistant
 from homeassistant.helpers import config_validation as cv
 
+from .config_flow import CONF_UPDATE_INTERVAL
 from .const import DEFAULT_UPDATE_INTERVAL, DOMAIN, IGNORED_TAHOMA_TYPES, TAHOMA_TYPES
 from .coordinator import TahomaDataUpdateCoordinator
 
--- a/custom_components/tahoma/__init__.py
+++ b/custom_components/tahoma/__init__.py
@@ -180,3 +180,5 @@ async def update_listener(hass, entry):
         coordinator.update_interval = (
             coordinator.original_update_interval
         ) = new_update_interval
+
+        await coordinator.async_refresh()
--- a/custom_components/tahoma/__init__.py
+++ b/custom_components/tahoma/__init__.py
@@ -1,6 +1,7 @@
 """The TaHoma integration."""
 import asyncio
 from collections import defaultdict
+from datetime import timedelta
 import logging
 
 from pyhoma.client import TahomaClient
@@ -9,6 +10,7 @@ import voluptuous as vol
 
 from homeassistant import config_entries
 from homeassistant.components.scene import DOMAIN as SCENE
+from homeassistant.components.tahoma.config_flow import CONF_UPDATE_INTERVAL
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import (
     CONF_EXCLUDE,
@@ -91,6 +93,8 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
         await client.close()
         return False
 
+    update_interval = entry.options.get(CONF_UPDATE_INTERVAL, DEFAULT_UPDATE_INTERVAL)
+
     tahoma_coordinator = TahomaDataUpdateCoordinator(
         hass,
         _LOGGER,
@@ -99,7 +103,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
         client=client,
         devices=await client.get_devices(),
         listener_id=await client.register_event_listener(),
-        update_interval=DEFAULT_UPDATE_INTERVAL,
+        update_interval=timedelta(seconds=update_interval),
     )
 
     await tahoma_coordinator.async_refresh()
--- a/custom_components/tahoma/__init__.py
+++ b/custom_components/tahoma/__init__.py
@@ -98,7 +98,6 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
     tahoma_coordinator = TahomaDataUpdateCoordinator(
         hass,
         _LOGGER,
-        # Name of the data. For logging purposes.
         name="TaHoma Event Fetcher",
         client=client,
         devices=await client.get_devices(),
--- a/custom_components/tahoma/__init__.py
+++ b/custom_components/tahoma/__init__.py
@@ -56,7 +56,7 @@ async def async_setup(hass: HomeAssistant, config: dict):
         configuration.get(CONF_USERNAME) in entry.data.get(CONF_USERNAME)
         for entry in hass.config_entries.async_entries(DOMAIN)
     ):
-        return False
+        return True
 
     hass.async_create_task(
         hass.config_entries.flow.async_init(
--- a/custom_components/tahoma/__init__.py
+++ b/custom_components/tahoma/__init__.py
@@ -12,12 +12,7 @@ import voluptuous as vol
 from homeassistant import config_entries
 from homeassistant.components.scene import DOMAIN as SCENE
 from homeassistant.config_entries import ConfigEntry
-from homeassistant.const import (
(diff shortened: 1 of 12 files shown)


custom_components/tahoma/__init__.py
```diff
@@ -56,7 +56,7 @@ async def async_setup(hass: HomeAssistant, config: dict):
         configuration.get(CONF_USERNAME) in entry.data.get(CONF_USERNAME)
         for entry in hass.config_entries.async_entries(DOMAIN)
     ):
-        return True
+        return False
```
iMicknl, line 59: This was a mistake, we are not setting up the entity here, we cancel it.

custom_components/tahoma/__init__.py
```diff
@@ -76,13 +76,25 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
...
     password = entry.data.get(CONF_PASSWORD)
 
+    client = TahomaClient(username, password)
+
     try:
-        client = TahomaClient(username, password)
         await client.login()
+    except TooManyRequestsException:
+        _LOGGER.error("too_many_requests")
+        await client.close()
```
iMicknl, line 85: We forgot to close the client on exceptions.. I also added logging, since this try/catch will be called on a restart for example.

custom_components/tahoma/const.py
```diff
@@ -9,6 +11,9 @@
 
 DOMAIN = "tahoma"
 
+MIN_UPDATE_INTERVAL = 1
+DEFAULT_UPDATE_INTERVAL = 30
```
iMicknl, line 15: Do we agree on 30 seconds as default? Or should we keep it 10?

custom_components/tahoma/strings.json
```diff
@@ -20,20 +20,13 @@
   },
   "options": {
     "step": {
-      "init": {
-        "description": "Select temperature sensors for thermostats and preset temperatures.\nNote: reset temperatures do not apply to the Smart Thermostat.",
+      "update_interval": {
+        "title": "Update Interval",
+        "description": "The Somfy TaHoma integration periodically checks for new events. Change the update interval to a lower value if you use a physical remote to control your devices.",
```
iMicknl, line 25: Could someone help me with the French translation? :-)

custom_components/tahoma/__init__.py
tetienne: If you return False I think it means an error occur during the setup.

custom_components/tahoma/__init__.py
```diff
@@ -76,13 +76,25 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
...
+        return False
+    except BadCredentialsException:
+        _LOGGER.error("invalid_auth")
```
tetienne, line 88: The log messages can be translated?

custom_components/tahoma/__init__.py
```diff
@@ -76,13 +76,25 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
...
+        await client.close()
+        return False
     except Exception as exception:  # pylint: disable=broad-except
         _LOGGER.exception(exception)
+        await client.close()
         return False
 
+    update_interval = entry.options.get(CONF_UPDATE_INTERVAL, DEFAULT_UPDATE_INTERVAL)
```
tetienne, line 96: [Here](https://github.com/iMicknl/ha-tahoma/pull/227/files#diff-7132e279b007f1584ef563c8b28cdd00R84) you set a default value, so you should be able to only do `entry.options[CONF_UPDATE_INTERVAL]`

custom_components/tahoma/config_flow.py
```diff
@@ -61,3 +70,37 @@ async def async_step_import(self, import_config: dict):
...
+
+    async def async_step_init(self, user_input=None):
+        """Manage the Somfy TaHoma options."""
+        return await self.async_step_update_interval()
+
+    async def async_step_update_interval(self, user_input=None):
+        """Manage the options regarding interval updates."""
+        if user_input is not None:
+            self.options[CONF_UPDATE_INTERVAL] = user_input[CONF_UPDATE_INTERVAL]
+            return self.async_create_entry(title="", data=self.options)
```
tetienne, line 94: You probably take this from dev documentation. Do you know if the empty title is actually store in the config entries file?

custom_components/tahoma/config_flow.py
```diff
@@ -61,3 +70,37 @@ async def async_step_import(self, import_config: dict):
...
+class OptionsFlowHandler(config_entries.OptionsFlow):
+    """Handle a option flow for TaHoma."""
+
+    def __init__(self, config_entry):
+        """Initialize options flow."""
+        self.config_entry = config_entry
+        self.options = dict(config_entry.options)
+
+        if self.options.get(CONF_UPDATE_INTERVAL) is None:
+            self.options[CONF_UPDATE_INTERVAL] = DEFAULT_UPDATE_INTERVAL
```
tetienne, line 84: Is it possible as using voluptuous you set a default value? https://github.com/iMicknl/ha-tahoma/pull/227/files#diff-7132e279b007f1584ef563c8b28cdd00R102

custom_components/tahoma/const.py
tetienne: I'm ok with this.

custom_components/tahoma/coordinator.py
```diff
@@ -50,16 +55,36 @@ async def _async_update_data(self) -> Dict[str, Device]:
             raise UpdateFailed(f"Error communicating with the TaHoma API: {exception}")
         else:
             for event in events:
+                if event.name == "DeviceAvailableEvent":
+                    self.devices[event.deviceurl].available = True
```
tetienne, line 59: When this event occurs?

custom_components/tahoma/coordinator.py
```diff
@@ -50,16 +55,36 @@ async def _async_update_data(self) -> Dict[str, Device]:
...
+
+                if (
+                    event.name == "ExecutionStateChangedEvent"
+                    and event.exec_id in self.executions
+                ) and event.new_state == "COMPLETED":
+                    self.executions.pop(event.exec_id, None)
+
+            if len(self.executions) < 1:
+                self.update_interval = self.original_update_interval
+
```
tetienne, line 82: What happen if I have both ExecutionRegisteredEvent and ExecutionStateChangedEvent at the same time? Did you check if it can occur with 30s poll interval for instance?

custom_components/tahoma/coordinator.py
```diff
@@ -38,8 +41,10 @@ def __init__(
             hass, logger, name=name, update_interval=update_interval,
         )
 
+        self.original_update_interval = update_interval
         self.client = client
         self.devices: Dict[str, Device] = {d.deviceurl: d for d in devices}
+        self.executions: Dict[str, defaultdict] = {}
```
tetienne, line 47:
```suggestion
        self.executions: Dict[str, Dict[str, str]] = {}
```

custom_components/tahoma/coordinator.py
```diff
@@ -50,16 +55,36 @@ async def _async_update_data(self) -> Dict[str, Device]:
...
                             device.states[state.name] = state
                         device.states[state.name].value = self._get_state(state)
+
+                if (
+                    event.name == "ExecutionRegisteredEvent"
+                    and event.exec_id in self.executions
+                ):
+                    self.update_interval = timedelta(seconds=1)
+
+                if (
```
tetienne, line 74: The braces are useless, you only use the and operator, their is no priority issue.

custom_components/tahoma/coordinator.py
```diff
@@ -50,16 +55,36 @@ async def _async_update_data(self) -> Dict[str, Device]:
...
+                    event.name == "ExecutionStateChangedEvent"
+                    and event.exec_id in self.executions
+                ) and event.new_state == "COMPLETED":
+                    self.executions.pop(event.exec_id, None)
+
+            if len(self.executions) < 1:
```
tetienne, line 80:
```suggestion
            if not self.executions):
```

```
>>> a = {}
>>> if not a:
...   print("Yeah")
... 
Yeah
>>> a = {"a1": "v1"}
>>> if not a:
...   print("Yeah")
... 
>>> 
```

custom_components/tahoma/coordinator.py
```diff
@@ -50,16 +55,36 @@ async def _async_update_data(self) -> Dict[str, Device]:
...
+                    event.name == "ExecutionRegisteredEvent"
+                    and event.exec_id in self.executions
+                ):
+                    self.update_interval = timedelta(seconds=1)
+
+                if (
+                    event.name == "ExecutionStateChangedEvent"
+                    and event.exec_id in self.executions
+                ) and event.new_state == "COMPLETED":
+                    self.executions.pop(event.exec_id, None)
```
tetienne, line 78:
```suggestion
                    del self.executions[event.exec_id]
```

```
>>> a = {"a1": "v1"}
>>> del a["a1"]
>>> a
{}
```

custom_components/tahoma/cover.py
```diff
@@ -229,6 +229,32 @@ async def async_stop_cover_tilt(self, **_):
...
 
+    @property
+    def is_opening(self):
+        """Return if the cover is opening or not."""
+        return next(
+            (
+                True
+                for execution in self.coordinator.executions.values()
+                if execution.get("deviceurl") == self.device.deviceurl
+                and execution.get("command_name") in list(COMMAND_OPEN, COMMAND_UP)
```
tetienne, line 240: `list(COMMAND_OPEN, COMMAND_UP)` -> ` [COMMAND_OPEN, COMMAND_UP]`

tetienne, line 240: I suggest you to use this syntax: `execution["deviceurl"]`. We know the value will be always present, and this is shorter.

custom_components/tahoma/strings.json
tetienne: I can help ;) But I don't understand your last sentence: 
>if you use a physical remote to control your devices.

I would suggest: 
> if you want more frequent update.

custom_components/tahoma/tahoma_device.py
```diff
@@ -152,6 +152,14 @@ def has_state(self, *states: str) -> bool:
...
             self.device.deviceurl, Command(command_name, list(args)), "Home Assistant"
         )
+
+        # ExecutionRegisteredEvent doesn't contain the deviceurl, thus we need to register it here
+        self.coordinator.executions[exec_id] = {
+            "deviceurl": self.device.deviceurl,
+            "command_name": command_name,
+        }
+
+        await self.coordinator.async_refresh()
```
tetienne, line 165: OK, I understand better the above code. Triggering an immediate update, you can catch the registered event.

custom_components/tahoma/coordinator.py
tetienne: We should perhaps also add a security in case we miss a COMPLETED event, in which case we will have 1s as poll interval forever.

custom_components/tahoma/config_flow.py
iMicknl: Unfortunately not, I did that first. However every time the OptionsFlow dialog opens, it will show the default value instead of the set value..

custom_components/tahoma/coordinator.py
iMicknl: When the device status changes back to available, for example if it was without batteries or outside the network.

custom_components/tahoma/coordinator.py
iMicknl: > What happen if I have both ExecutionRegisteredEvent and ExecutionStateChangedEvent at the same time? Did you check if it can occur with 30s poll interval for instance?

Does it matter? If there are no executions left, the update interval will be set to the default interval in the end. I wrote this with a high interval in mind.

custom_components/tahoma/coordinator.py
iMicknl: Indeed. Firstly I didn't check here `if event.exec_id in self.executions`, thus I could not use `del`. However since we check it up front, we could use `del()`.

custom_components/tahoma/__init__.py
iMicknl: No.. This was a placeholder for now, I think we need to pass the original exception messages.

custom_components/tahoma/coordinator.py
iMicknl: The braces are added by black.. I believe because it is a multi line if statement, due to the amount of characters.

custom_components/tahoma/__init__.py
iMicknl: I am still figuring this out. The Option Flow is triggered only when you press `options` now, thus the value can be empty.

custom_components/tahoma/strings.json
iMicknl: >The Somfy TaHoma integration periodically retrieves new events. Change the update interval to a lower value if you if you want more frequent updates, for example when you also control your devices outside Home Assistant.

Good one, this is the latest text, does it make more sense? I think it would be good to mention when you need a higher update frequency, since it is not necessary if you only use Home Assistant.

custom_components/tahoma/config_flow.py
iMicknl: Yes, but all examples I have seen (like Hue) are using an empty title. It looks like it is not used.

custom_components/tahoma/__init__.py
iMicknl: @tetienne @vlebourl what do you think? I am fine by having those short descriptive messages since this is part of the logging.

We should maybe reinit the config flow when we get a `BadCredentialsException`? Not sure how other components do that.

tetienne: In such case, I think the return False is enough. The user would have to remove and re add our integration to change its login/password. In can only occurs when it changes it itself.

custom_components/tahoma/cover.py
```diff
@@ -229,6 +229,34 @@ async def async_stop_cover_tilt(self, **_):
             self.select_command(COMMAND_STOP_IDENTIFY, COMMAND_STOP, COMMAND_MY)
         )
 
+    @property
+    def is_opening(self):
+        """Return if the cover is opening or not."""
+        return next(
```
tetienne, line 235: Have a look at the any function to ease your code.

custom_components/tahoma/coordinator.py
```diff
@@ -50,16 +57,42 @@ async def _async_update_data(self) -> Dict[str, Device]:
...
+
+                if (
+                    event.name == "ExecutionRegisteredEvent"
+                    and event.exec_id in self.executions
+                ):
+                    self.update_interval = timedelta(seconds=1)
+
+                if (
+                    event.name == "ExecutionStateChangedEvent"
+                    and event.exec_id in self.executions
```
tetienne, line 82: I don't what to think about this. In our tahoma library, I now update the device state and recreate the session to solve the disconnect session error we met: https://github.com/iMicknl/python-tahoma-api/pull/36/files#diff-f6a3053f6cfdb465b9897b1a29772ea4R96
The linked code will be not compatible with your's :S

vlebourl: To be honest, I'd rather keep the update at a high frequency all the time. I use the tahoma app quite often and I'd like the state to be reflected promptly in HA rather than wait for an update. Plus according to the tahoma doc on event listener, it seems to be built to have fetches every seconds. IMHO having the reconnect in case of a disconnection is a better solution.

iMicknl: @vlebourl in that case you are able to set the event listener to 1 second via the options? I can understand that there are users who are not comfortable with having a request every second. And I am not sure if that will even be accepted in core, as a default...

This change will make sure that users are able to configure the default polling interval, thus they can choose to instant (every second) and 'controlled by HA only' and set it to an hour or more. However, in case of a long polling interval (like our current default), updates will be very slow after taking an action from HA, which should not be normal behaviour in my opinion.

vlebourl: fair enough. Is there a way to reconcile @tetienne 's code though?

iMicknl: > I don't what to think about this. In our tahoma library, I now update the device state and recreate the session to solve the disconnect session error we met: [iMicknl/python-tahoma-api#36 (files)](https://github.com/iMicknl/python-tahoma-api/pull/36/files#diff-f6a3053f6cfdb465b9897b1a29772ea4R96)
> The linked code will be not compatible with your's :S

Let's not merge this PR until we figure out the best way! I can understand your approach, however I believe this will make the package less flexible. My preference would be to try to auto reconnect and throw a certain status, so that we can retrieve the latest device states with a force refresh in the HA integration. 

Did we try to see if our current event listener can still be retrieved, after an error? We could re login automatically and see if we can pull the same event listener.

vlebourl: I like that approach

tetienne: I didn't try to recall the same listener once logout and re-login. But's it's safer to recreate it.

Look at the Hue integration, they set the poll interval to 5 seconds: https://github.com/home-assistant/core/blob/dev/homeassistant/components/hue/light.py#L38

So we should be able to do the same.

custom_components/tahoma/coordinator.py
```diff
@@ -49,16 +46,15 @@ def __init__(
...
     async def _async_update_data(self) -> Dict[str, Device]:
         """Fetch TaHoma data via event listener."""
         try:
-            events = await self.client.fetch_event_listener(self.listener_id)
+            events = await self.client.fetch_events()
         except (ServerDisconnectedError, NotAuthenticatedException) as exception:
             _LOGGER.debug(exception)
             await self.client.login()
-            self.listener_id = await self.client.register_event_listener()
+            await self.client.register_event_listener()
```
tetienne, line 57: You can remove this line now. We do it in the login.

custom_components/tahoma/coordinator.py
```diff
@@ -62,17 +62,45 @@ async def _async_update_data(self) -> Dict[str, Device]:
...
+                event.name == "ExecutionRegisteredEvent"
+                and event.exec_id in self.executions
+            ):
+                self.update_interval = timedelta(seconds=1)
+
+            if (
+                event.name == "ExecutionStateChangedEvent"
+                and event.exec_id in self.executions
+                and event.new_state == "COMPLETED"
+                or event.new_state == "FAILED"
```
tetienne, line 92: I think you have a priority issue.

```suggestion
                event.name == "ExecutionStateChangedEvent"
                and event.exec_id in self.executions
                and (event.new_state == "COMPLETED"
                or event.new_state == "FAILED")
```

custom_components/tahoma/coordinator.py
```diff
@@ -62,17 +62,45 @@ async def _async_update_data(self) -> Dict[str, Device]:
...
+            elif event.name == "DeviceStateChangedEvent":
                 for state in event.device_states:
                     device = self.devices[event.deviceurl]
                     if state.name not in device.states:
                         device.states[state.name] = state
                     device.states[state.name].value = self._get_state(state)
 
+            if (
+                event.name == "ExecutionRegisteredEvent"
+                and event.exec_id in self.executions
```
tetienne, line 84: Finally, do we care if the command has been triggered by HA? Imagine I launch it from the tahoma website, it will be nice to have the update asap in HA too.

tetienne (changes requested):

custom_components/tahoma/coordinator.py
iMicknl: True! I made the changes for this, but we would need to test it.

custom_components/tahoma/config_flow.py
```diff
@@ -77,3 +90,37 @@ async def async_step_import(self, import_config: dict):
...
```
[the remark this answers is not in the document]
iMicknl: Not sure how to simplify, any suggestion? Do you mean that the dict(config_entry.options) could be removed? I could not find a lot of documentation about the OptionsFlow, so I had a look at some other components for this.

vlebourl (approved):

tetienne (approved):

Issue #223, opened 2020-08-12
[Architecture] Allow multiple entities per device


iMicknl:
**Is your feature request related to a problem? Please describe.**
Currently, we only have one entity per discovered Somfy TaHoma device. However, looking at some of the devices, there are more interesting states and/or actions to take, which we put now in attributes.

Looking at PR's to core, I get the feeling that the attributes part is used less and less and should not be used to store everything. 

For example, looking at some sensors, there are multiple power states and the same for some security devices.

**Describe the solution you'd like**
Rewrite part of the integration to dynamically allow the creation of multiple entities per device.

**Additional context**
Example of a Z-Wave device, with multiple entities. Some less frequent used entities are hidden by default. 
<img width="354" alt="Screenshot 2020-08-12 at 17 44 53" src="https://user-images.githubusercontent.com/1424596/90036612-8b2df600-dcc3-11ea-9fd5-5f99608dc4f0.png">

tetienne:
So you want to convert some Somfy attribute to hass entity_id?

iMicknl:
@tetienne not sure yet how to do it technically. Opened this issue for tracking, since I faced it a few times in my last work on the integration.

Open for any technical architectures.

PoVPoV29:
It should be perfect to manage detectors ! :
![Screenshot_20200813-114556_Somfy Protect](https://user-images.githubusercontent.com/69581008/90120358-f1198c80-dd5a-11ea-8ac6-ab2a257798ae.jpg)

PoVPoV29:
It's should be perfect to manager detector : 
![Capture](https://user-images.githubusercontent.com/69581008/90120899-959bce80-dd5b-11ea-84e9-5b810b883498.PNG)

iMicknl:
I started on this one, however there is a lot of work to be done.

Sneak peek:
<img width="333" alt="Screen Shot 2021-04-18 at 18 31 39" src="https://user-images.githubusercontent.com/1424596/115153078-55bc8200-a074-11eb-95a6-75c9488365ab.png">

iMicknl:
@Hankey71 the version below should add a sensor for the expected number of showers.
@tillstaff the version below should add a sensor for the speed of each motors (inlet and outlet). 

Could you give https://github.com/iMicknl/ha-tahoma/archive/refs/heads/feature/add_secondary_sensors.zip a try? Extract this file and place `custom_components/tahoma` in your `custom_components` folder. Let me know how it works.

[...]

Issue #233, opened 2020-08-18
Cover position changes made from the wall switch are never reflected in HA


romain38:
**Describe the bug**

Cover position changes made from the wall switch are never reflected in Home-Assistant.
If I run mobile app Tahoma, or run stop command from Home-Assistant, right position are updated.
I've try to call service `homeassistant.update_entity` on my cover sensor without effect.

**To Reproduce**

Force refresh with stop sensor command:
- Cover opened at 100% (physically and in Home-Assistant)
- Close cover via wall switch (Somfy IO remote control: SMOOVE RS100 iO)
- Cover physically closed but Home-Assistant keep opened status at 100%
- Wait some minutes, Home-Assistant keep opened status still at 100%
- Call cover.stop_cover command on cover sensor in Home-Assistant
- Home-Assistant update cover status to closed

Force refresh with Tahoma android app:
- Cover opened at 100% (physically and in Home-Assistant)
- Close cover via wall switch (Somfy IO remote control: SMOOVE RS100 iO)
- Cover physically closed but Home-Assistant keep opened status at 100%
- Wait some minutes, Home-Assistant keep opened status still at 100%
- Run Tahoma android app, cover is displayed as closed in Tahoma mobile app
- Home-Assistant update cover status to closed

**Expected behavior**

Cover status must automatically reflect the right cover position (with an acceptable refresh delay) without call manually stop command or use Tahoma AndroidApp.

**Environment (please complete the following information):**

- **ha-tahoma version:** 2.0.1
- **Home Assistant version:** 0.112.4
- **Platform:** cover
- **Firmware:** io:RollerShutterWithLowSpeedManagementIOComponent

**Devices:**

Somfy IO cover
Wall switch SMOOVE RS100 IO
Tahoma Box

iMicknl:
I am afraid this is due to #167 as well. Did you not have the same behavior with the old version in Home Assistant core?

Could you share some more details about your device? Settings -> Integrations -> Devices -> (click your cover device. Could you share the firmware?

Example:
<img width="347" alt="Screenshot 2020-08-18 at 10 55 26" src="https://user-images.githubusercontent.com/1424596/90492314-5990b100-e141-11ea-834f-b6424d2cce7a.png">

romain38:
Hi, thank you for this quick response! 

[...]

Issue #230, opened 2020-08-14
No state update for StatefulOnOffLight (eliot:OnOffLightEliotComponent)


HaSHi93110:
**Describe the bug**
hi.
I have test your integration.
I have in my tahoma box a legrand light which is linked from legrand account.
In the tahoma application the state of my light works good. When i turn on the light in my legrand application, the state in the tahoma application refresh and show me on. But in ha, i can't see the state on of my light. The state is always off and i can't change manually the state.

**To Reproduce**

1. Go to service in developer tools
2. Launch light.turn_on in the light entity
3. The light is on
4. The state of the light is always off

**Expected behavior**
When I turn on the light, the state of the entity should be "on" and no "off"

**Environment (please complete the following information):**
- **ha-tahoma version:** 2.0.1
- **Home Assistant version:** 0.113.3
- **Platform**: light

**Device:** interrupteur legrand celiane with netatmo (with neutral) linked in tahoma application

- **Model**: StatefulOnOffLight
- **Type**: eliot:OnOffLightEliotComponent

**Additional context**

```
{
	"commands": [{
		"commandName": "off",
		"nparams": 0
	}, {
		"commandName": "on",
		"nparams": 0
	}, {
		"commandName": "onWithTimer",
		"nparams": 1
	}, {
		"commandName": "refreshState",
		"nparams": 0
	}, {
		"commandName": "setOnOff",
		"nparams": 1
	}, {
		"commandName": "refreshAllLightStates",
		"nparams": 0
	}],
	"states": [{
		"type": "DataState",
		"qualifiedName": "core:FirmwareRevisionState"
	}, {
		"values": ["off", "on"],
		"type": "DiscreteState",
		"qualifiedName": "core:OnOffState"
	}],
	"dataProperties": [],
	"widgetName": "StatefulOnOffLight",
	"uiClass": "Light",
	"qualifiedName": "eliot:OnOffLightEliotComponent",
	"type": "ACTUATOR"
}
```

iMicknl:
Thanks @HaSHi93110! I haven't seen an `eliot:` device (legrand) before. It looks related to #167, since there is a command available called `refreshAllLightStates`.

tetienne:
It would be interesting to dump the tahoma website to see where in their code they use this command. @iMicknl What do you think?

HaSHi93110:
When i turn on the light with tahoma app, thé light appear on in 5 6 seconds in the app, and she is also on in ha.
[...]

PR #241, opened 2020-08-30
Add service to refresh states on TaHoma box

### LINKED ISSUE #167: Implement optional update coordinator to call `refreshState` periodically
When we move towards using an event based approach for updating states (#1), it would be good to understand if we need to refresh the states manually for specific devices, or if we can do it in batch.

For example;
- `io:DiscreteGateOpenerIOComponent` uses `refreshPedestrianPosition` (#40)
- `io:SimpleExteriorHeatingIOComponent` uses `getLevel` possibly (#161)
and there are more examples
- `eliot:OnOffLightEliotComponent` uses `refreshAllLightStates` (#230)

Getting rid of manually calling all refresh endpoints would be best, or scheduling a refresh of all states periodically. 

**Describe alternatives you've considered**
Maintaining a list of states that have to be refreshed manually.

iMicknl:
A first try to solve the issues around states for devices that need a refresh. In this PR a service is added to manually refresh the states. If this is the solution, we could add this as an option to the DataUpdateCoordinator.

Fixes #167, #233, #230

Please note that after the refresh is complete, the event needs to be fetched via the DataUpdateCoordinator.
```2020-08-30 12:08:58 DEBUG (MainThread) [custom_components.tahoma.coordinator] RefreshAllDevicesStatesCompletedEvent/None (device:None,state:None->None)```

### DIFF
--- a/custom_components/tahoma/__init__.py
+++ b/custom_components/tahoma/__init__.py
@@ -7,6 +7,7 @@ import logging
 from aiohttp import CookieJar
 from pyhoma.client import TahomaClient
 from pyhoma.exceptions import BadCredentialsException, TooManyRequestsException
+from pyhoma.models import Command
 import voluptuous as vol
 
 from homeassistant import config_entries
@@ -31,6 +32,8 @@ from .const import (
 from .coordinator import TahomaDataUpdateCoordinator
 
 _LOGGER = logging.getLogger(__name__)
+
+SERVICE_EXECUTE_COMMAND = "execute_command"
 
 CONFIG_SCHEMA = vol.Schema(
     {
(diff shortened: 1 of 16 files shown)


tetienne:
Why not create an other coordinator that will call regularly the refresh request?

iMicknl:
> Why not create an other coordinator that will call regularly the refresh request?

That would be my next step. This was just to test if this command works, since I cannot test it myself unfortunately.

vlebourl:
How would we decide the frequency of those refresh? We have to profile that code before making it automatic to be sure it doesn't consume too much resources.

vlebourl:
How does this fix  #223?

iMicknl:
> How does this fix #223?

#233 😂. Apparently I only fixed this link on the 'linked issues' part, but forgot to update my original post.

iMicknl:
> How would we decide the frequency of those refresh? We have to profile that code before making it automatic to be sure it doesn't consume too much resources.

Not sure yet. In HomeBridge they make this configurable and I saw many people putting it pretty low (60 seconds).

PR #242, opened 2020-08-31
Add service to set cover position with low speed

iMicknl:
Verified by user clemalex on the HA Community forum.

### DIFF
--- a/custom_components/tahoma/cover.py
+++ b/custom_components/tahoma/cover.py
@@ -72,6 +72,10 @@ STATE_CLOSED = "closed"
 SERVICE_MY = "cover_my"
 SERVICE_COVER_POSITION_LOW_SPEED = "set_cover_position_low_speed"
 
+SUPPORT_MY = 512
+SUPPORT_COVER_POSITION_LOW_SPEED = 1024
+
+
 TAHOMA_COVER_DEVICE_CLASSES = {
     "Awning": DEVICE_CLASS_AWNING,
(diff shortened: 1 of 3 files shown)


custom_components/tahoma/cover.py
```diff
@@ -103,6 +107,16 @@ async def async_setup_entry(hass, entry, async_add_entities):
         SERVICE_MY, {}, "async_my",
     )
 
+    platform.async_register_entity_service(
```
vlebourl, line 110: is there a way to only register the service if the command exists?

iMicknl: Not that I am aware of. And even if this would be possible, you would still register it for all covers unfortunately. I figured out that you can pass a `SUPPORT_XXX` when you register the service, but this only works for predefined `SUPPORTED_FEATURES` I believe..

vlebourl: fair enough

vlebourl (approved):

custom_components/tahoma/cover.py
tetienne: You can loop on all the devices and check if at least one has the command.

iMicknl: @tetienne I did implement some extra `SUPPORTED_FEATURES` for now. This looks for me like the cleanest 'core' way. (if they accept this .. ;-))

iMicknl:
Confirmed by clemalex that the latest version still works. I would be happy to merge it after review.

Issue #211, opened 2020-08-07
Add support for Modulis Heater (rts:ExteriorHeatingRTSComponent)


larsvinc:
**Device information**
The device is a terrasse heater:
https://www.somfy.me/en-gc/products/1810876/heating-modulis-ramp-rts-

It uses the same buttons as my blinds:
- "up"/"on" (on for the heater)
- "down"/"off" (off for the heater)
- "my" Programmable setting for the heater

If it exposes more functions through the API that would be great to have them supported (such as selecting the heating level).

**Device details**

vlebourl:
```
2020-08-06 18:22:01 DEBUG (MainThread) [custom_components.tahoma] Unsupported TaHoma device detected (rts:ExteriorHeatingRTSComponent - ExteriorHeatingSystem - StatelessExteriorHeating)
```

iMicknl:
```
{
	"commands": [{
		"commandName": "down",
		"nparams": 0
	}, {
		"commandName": "identify",
		"nparams": 0
	}, {
		"commandName": "my",
		"nparams": 0
	}, {
		"commandName": "off",
		"nparams": 0
	}, {
		"commandName": "on",
		"nparams": 0
	}, {
		"commandName": "onWithTimer",
		"nparams": 1
	}, {
		"commandName": "rest",
		"nparams": 0
	}, {
		"commandName": "stop",
		"nparams": 0
	}, {
		"commandName": "test",
		"nparams": 0
	}, {
		"commandName": "up",
		"nparams": 0
	}],
	"states": [],
	"dataProperties": [{
		"value": "0",
		"qualifiedName": "core:identifyInterval"
	}],
	"widgetName": "StatelessExteriorHeating",
	"uiClass": "ExteriorHeatingSystem",
	"qualifiedName": "rts:ExteriorHeatingRTSComponent",
	"type": "ACTUATOR"
}
```

larsvinc:
Some comments regarding wanted functionality:
- I suspect that "up" and "down" differs from "on" and "off". It is possible to adjust the heating with "up" and "down" after the unit is turned on. So it would make sense to enable all those. 
- "my" sets the heating at a pre-programmed level
- As for "rest" and "stop", I have no idea what they do. Is it possible for me to test the full API and report back what the commands do?

iMicknl:
@vlebourl do you think it is possible to map this one to the `climate` platform as well? 

[...]

PR #244, opened 2020-08-31
Add support for StatelessExteriorHeating device.

vlebourl:
implement #211


custom_components/tahoma/climate_seh.py
```diff
@@ -0,0 +1,68 @@
...
+
+    @property
+    def temperature_unit(self) -> str:
+        """Return the unit of measurement used by the platform."""
+        return TEMP_CELSIUS
+
+    @property
+    def hvac_mode(self) -> str:
+        """Return hvac operation ie. heat, cool mode."""
+        return HVAC_MODE_UNKNOWN
```
iMicknl, line 44: What if you return `None`? Normally you don't return a string, but `None` will default to `Unknown`.

custom_components/tahoma/climate_seh.py
```diff
@@ -0,0 +1,68 @@
...
+
+    @property
+    def supported_features(self) -> int:
+        """Return the list of supported features."""
+        return 0
+
+    @property
+    def temperature_unit(self) -> str:
+        """Return the unit of measurement used by the platform."""
+        return TEMP_CELSIUS
```
iMicknl, line 39: Is this measured in Celsius?

custom_components/tahoma/climate_aeh.py
```diff
@@ -87,3 +90,15 @@ async def async_set_preset_mode(self, preset_mode: str) -> None:
     async def async_turn_off(self) -> None:
         """Turn off the device."""
         await self.async_execute_command(COMMAND_SET_HEATING_LEVEL, PRESET_OFF)
+
+    async def async_down(self, **_):
```
iMicknl, line 94: Is this required in all climate integrations?

custom_components/tahoma/climate_seh.py
```diff
@@ -0,0 +1,68 @@
...
+    """Representation of TaHoma IO Atlantic Electrical Heater."""
+
+    def __init__(self, tahoma_device, controller):
+        """Init method."""
+        super().__init__(tahoma_device, controller)
+
+    @property
+    def supported_features(self) -> int:
+        """Return the list of supported features."""
+        return 0
```
iMicknl, line 34: Could we do something with `SUPPORT_FAN_MODE` or `SUPPORT_AUX_HEAT`? Not sure if there are any possibilities to set a preset (up + down). Since this is a RTS device, we will set `assumed_state` to True already, thus I believe both options should be active all the time.

Or using `SUPPORT_PRESET_MODE`?

custom_components/tahoma/climate_seh.py
vlebourl: I did this to have a `str` but if `None` is fine, that would be better!

custom_components/tahoma/climate_aeh.py
vlebourl: I don't know what happens if you call a service on an entity that doesn't have the method.

custom_components/tahoma/climate_seh.py
vlebourl: There's no temperature in that device but `temperature_unit` is mandatory... I guess `None` would be fine as well.

custom_components/tahoma/climate_seh.py
vlebourl: I first started with SUPPORT_PRESET_MODE with PRESET_ECO, PRESET_COMFORT and PRESET_BOOST as the three levels. But there is no way to know what preset is selected, so how many ups and downs should be done to go to the desired preset. One solution could be 2 ups for boost, 2 downs for Eco, and 2 ups 1 down or 2 downs 1 up for comfort. It's a bit overkill, but if it doesn't cycle through the modes, that could work.

custom_components/tahoma/climate_seh.py
vlebourl: changed to None

custom_components/tahoma/climate_seh.py
vlebourl: changed to None

custom_components/tahoma/climate_seh.py
vlebourl: added presets.

custom_components/tahoma/climate.py
```diff
@@ -1,18 +1,27 @@
...
+_LOGGER = logging.getLogger(__name__)
+
 TYPE = {
     "AtlanticElectricalHeater": AtlanticElectricalHeater,
     "SomfyThermostat": SomfyThermostat,
     "DimmerExteriorHeating": DimmerExteriorHeating,
+    "StatelessExteriorHeating": StatelessExteriorHeating,
 }
 
+SERVICE_MY = "climate_my"
```
iMicknl, line 22:
```suggestion
SERVICE_CLIMATE_MY_POSITION = "set_climate_my_position"
```

Reminder to eventually align this to #242, if we al agree on a final name.

custom_components/tahoma/climate_devices/stateless_exterior_heating.py
```diff
@@ -0,0 +1,97 @@
+"""Support for Atlantic Electrical Heater IO controller."""
+import logging
+from pprint import pformat
+from typing import List, Optional
+
+from custom_components.tahoma.climate import SUPPORT_MY
```
iMicknl, line 6: This will not work in all situations I believe. Could you use the notation without `custom-components`.

vlebourl: you're write! I let this slip through my checks...

custom_components/tahoma/climate.py
```diff
@@ -1,30 +1,48 @@
...
+
 TYPE = {
     "AtlanticElectricalHeater": AtlanticElectricalHeater,
     "SomfyThermostat": SomfyThermostat,
     "DimmerExteriorHeating": DimmerExteriorHeating,
+    "StatelessExteriorHeating": StatelessExteriorHeating,
 }
 
+SERVICE_CLIMATE_MY_POSITION = "set_climate_my_position"
+SUPPORT_MY = pow(2, 20)
```
iMicknl, line 24: Should we just default to the number value? I believe this function call is not necessary and doesn't improve readability. (since the other ones are specified as a number, thus I would need to now the result of pow(2, 20) to figure out if there are duplicates.

custom_components/tahoma/climate_devices/stateless_exterior_heating.py
```diff
@@ -0,0 +1,93 @@
+"""Support for Atlantic Electrical Heater IO controller."""
```
iMicknl, line 1:
```suggestion
"""Support for Stateless Exterior Heating."""
```

custom_components/tahoma/climate_devices/stateless_exterior_heating.py
```diff
@@ -0,0 +1,93 @@
...
+
+_LOGGER = logging.getLogger(__name__)
+
+COMMAND_DOWN = "down"
+COMMAND_MY = "my"
+COMMAND_OFF = "off"
+COMMAND_ON = "on"
+COMMAND_UP = "up"
+
+SUPPORT_MY = pow(2, 20)
```
iMicknl, line 26: Same as above

custom_components/tahoma/climate_devices/stateless_exterior_heating.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class StatelessExteriorHeating(TahomaDevice, ClimateEntity):
+    """Representation of TaHoma Stateless Exterior Heating device."""
+
+    def __init__(self, tahoma_device, controller):
```
iMicknl, line 32: For debug purposes I guess, can be removed afterwards. What you pass here by the way `device_url` followed by `coordinator`, not the device and controller.

vlebourl: will be removed once working!

tetienne (approved):

custom_components/tahoma/climate_devices/stateless_exterior_heating.py
```diff
@@ -0,0 +1,73 @@
...
+    HVAC_MODE_OFF,
+    SUPPORT_PRESET_MODE,
+)
+from homeassistant.const import TEMP_CELSIUS
+
+from ..tahoma_device import TahomaDevice
+
+_LOGGER = logging.getLogger(__name__)
+
+COMMAND_DOWN = "down"
```
iMicknl, line 17: Not used

custom_components/tahoma/climate_devices/stateless_exterior_heating.py
```diff
@@ -0,0 +1,73 @@
...
+COMMAND_MY = "my"
+COMMAND_OFF = "off"
+COMMAND_ON = "on"
+COMMAND_UP = "up"
```
iMicknl, line 21: Not used

PR #250, opened 2020-09-01
Moved climate devices files to their own directory.

vlebourl:


tetienne (changes requested):
The name of your modules must be in snake_case.

tetienne (approved):

PR #251, opened 2020-09-02
Improve `stop` and `stop tilt` command for covers

iMicknl:


custom_components/tahoma/cover.py
```diff
@@ -258,16 +258,58 @@ async def async_close_cover_tilt(self, **_):
...
+            None,
         )
 
+        if exec_id:
+            logging.debug("Cancelling command " + exec_id)
+            await self.async_cancel_command(exec_id)
+        else:
+            logging.debug("Calling stop command")
+            await self.async_execute_command(
+                self.select_command(COMMAND_STOP_IDENTIFY, COMMAND_STOP, COMMAND_MY)
```
tetienne, line 310: You can create a stop method, to avoid duplicate code.

custom_components/tahoma/cover.py
```diff
@@ -258,16 +258,58 @@ async def async_close_cover_tilt(self, **_):
...
+        if exec_id:
+            logging.debug("Cancelling command " + exec_id)
+            await self.async_cancel_command(exec_id)
+        else:
+            logging.debug("Calling stop command")
+            await self.async_execute_command(
+                self.select_command(COMMAND_STOP, COMMAND_STOP_IDENTIFY, COMMAND_MY)
+            )
+
     async def async_stop_cover_tilt(self, **_):
```
tetienne, line 291: This method is exactly the same that the above one, except for the command names, you can create a common method.

custom_components/tahoma/cover.py
iMicknl: @tetienne true, there is quite some code duplication in `cover.py`. I will try to tackle some of it in this PR.

I just pushed my latest changes, which are not working yet, since I still need to fix the `select_command` to be able to handle a list.

iMicknl: Fixed.

custom_components/tahoma/cover.py
```diff
@@ -238,36 +250,54 @@ def icon(self):
...
+                for exec_id, execution in self.coordinator.executions.items()
+                if execution.get("deviceurl") == self.device.deviceurl
+                and execution.get("command_name") in cancel_commands
+            ),
+            None,
+        )
+
+        # Cancelling a running execution will stop the cover movement
+        if exec_id:
+            logging.debug("Cancelling command " + exec_id)
```
iMicknl, line 294: TODO remove when not in draft anymore

custom_components/tahoma/cover.py
```diff
@@ -238,36 +250,54 @@ def icon(self):
...
-            self.select_command(COMMAND_STOP_IDENTIFY, COMMAND_STOP, COMMAND_MY)
+        """Stop the cover tilt."""
+        await self.async_cancel_or_stop_cover(
+            COMMANDS_OPEN_TILT + COMMANDS_SET_TILT_POSITION + COMMANDS_CLOSE_TILT,
+            COMMANDS_STOP_TILT,
         )
 
+    async def async_cancel_or_stop_cover(self, cancel_commands, stop_commands) -> None:
+        """Cancel running execution or send stop command."""
+        exec_id = next(
```
iMicknl, line 282: @tetienne @vlebourl I just realised, maybe we should take the last command from the executions? (or reverse the array first and take next).

What if I open it and than press close. Close will cancel the first open, but if they are both still in the queue, it will be trying to cancel the wrong one perhaps.

custom_components/tahoma/cover.py
```diff
@@ -238,36 +250,54 @@ def icon(self):
...
+            (
+                exec_id
+                for exec_id, execution in self.coordinator.executions.items()
```
iMicknl, line 285:
```suggestion
                for exec_id, execution in self.coordinator.executions.items()
```
```suggestion
                for exec_id, execution in self.coordinator.executions.reverse().items()
```

Does this make sense? Or any better way to solve this.

iMicknl: .reverse doesn't work on a dict. Changed it to reversed(), which only works in Python 3.8 for Dicts.

custom_components/tahoma/cover.py
iMicknl: Added reverse functionality for the Dict. It is an edge case, but I could reproduce a situation where the first call of stop didn't work.

custom_components/tahoma/cover.py
```diff
@@ -47,6 +47,19 @@
 COMMAND_STOP_IDENTIFY = "stopIdentify"
 COMMAND_UP = "up"
 
+COMMANDS_STOP = [COMMAND_STOP_IDENTIFY, COMMAND_STOP, COMMAND_MY]
+COMMANDS_STOP_TILT = [COMMAND_STOP_IDENTIFY, COMMAND_STOP, COMMAND_MY]
```
tetienne, line 51: Do you think  later these 2 lists will be different? Can we use only one for the moment?

custom_components/tahoma/cover.py
```diff
@@ -238,36 +250,52 @@ def icon(self):
...
+            COMMANDS_OPEN_TILT + COMMANDS_SET_TILT_POSITION + COMMANDS_CLOSE_TILT,
+            COMMANDS_STOP_TILT,
         )
 
+    async def async_cancel_or_stop_cover(self, cancel_commands, stop_commands) -> None:
+        """Cancel running execution or send stop command."""
+        exec_id = next(
+            (
+                exec_id
+                for exec_id, execution in reversed(self.coordinator.executions.items())
```
tetienne, line 285: Why reverse the iterator?

custom_components/tahoma/cover.py
iMicknl: Not 100% sure yet... For the devices with tilt position that I have tested, the stop command didn't work, while the stop identify did. However, with this PR the stop command works based on the cancel method, if the command was executed from Home Assistant.

Another idea I had was to pull the running executions via the API; when you call stop without any running executions. But that would be for a next PR.

custom_components/tahoma/cover.py
iMicknl: To fix the edge case where you open an directly close a cover. If you try to stop the command, it still has the execution_id from `open` and `close`. We need to cancel the last command, thus this is why I reversed the iterator.

tetienne: OK, can you add a comment please? In few months/weeks, we will forget about this.

iMicknl: Added.

vlebourl (approved):

tetienne (approved):

PR #253, opened 2020-09-04
Bump pyhoma to 0.4.2

vlebourl:


iMicknl (commented):
Best would be to also update the version in the requirements.test.txt. And we don’t really need this in a separate PR, since it is not adding value without other modifications. We could wait until #251 is merged :-)

vlebourl:
Ok, i'll let you add it to your pr then

vlebourl:
Ok, i'll let you add it to your pr then

PR #263, opened 2020-09-14
Support cancelling cover movement when initiated outside Home Assistant

iMicknl:
(not happy with the readability of the code yet, thus marking it as a draft)

### DIFF
--- a/custom_components/tahoma/cover.py
+++ b/custom_components/tahoma/cover.py
@@ -47,8 +47,8 @@ COMMAND_STOP = "stop"
 COMMAND_STOP_IDENTIFY = "stopIdentify"
 COMMAND_UP = "up"
 
-COMMANDS_STOP = [COMMAND_STOP_IDENTIFY, COMMAND_STOP, COMMAND_MY]
-COMMANDS_STOP_TILT = [COMMAND_STOP_IDENTIFY, COMMAND_STOP, COMMAND_MY]
+COMMANDS_STOP = [COMMAND_STOP, COMMAND_STOP_IDENTIFY, COMMAND_MY]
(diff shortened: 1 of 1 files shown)


iMicknl:
Functionality works, as confirmed [here](https://github.com/iMicknl/ha-tahoma/issues/283#issuecomment-703302427).

custom_components/tahoma/cover.py
```diff
@@ -292,10 +292,27 @@ async def async_cancel_or_stop_cover(self, cancel_commands, stop_commands) -> No
...
+                and command.get("name") in cancel_commands
+            ),
+            None,
+        )
+
+        if exec_id:
+            return await self.async_cancel_command(exec_id)
+
+        # Fallback to available stop commands when no execution can be found
+        await self.async_execute_command(self.select_command(*stop_commands))
```
tetienne, line 315: Finally, why not keep only this one? Do you think some device does not have it?

iMicknl: You mean, removing the other logic in favour of this one? Unfortunately the stop command don't work for every device. Even when a device has a stop command (like my pergola), it just doesn't work.

Cancelling a running execution is the method to stop a cover movement (also on tahomalink.com) and it is a reliable way that works with every device. This is some Somfy logic that I couldn't work around unfortunately and it was causing quite some user issues.

vlebourl (approved):

custom_components/tahoma/cover.py
iMicknl: @tetienne good point by the way. I updated the documentation to make this design choice more clear.

tetienne (approved):

PR #271, opened 2020-09-18
Add service to refresh state periodically

### LINKED ISSUE #167: Implement optional update coordinator to call `refreshState` periodically

iMicknl:
Screenshot 2020-09-19 at 01 21 55

Adds support for calling `/setup/devices/states/refresh` periodically, using the existing data update coordinator.

Fixes #167.


custom_components/tahoma/__init__.py
```diff
@@ -198,3 +229,6 @@ async def update_listener(hass: HomeAssistant, entry: ConfigEntry):
         coordinator.original_update_interval = new_update_interval
 
         await coordinator.async_refresh()
+
+    if entry.options[CONF_REFRESH_STATE_INTERVAL]:
+        print("Change refresh state of async_track_time_interval")
```
iMicknl, line 234: @tetienne would you have a clue how to solve this? `async_track_time_interval` returns an unsub function, however I don't think I can create a new task calling `handle_refresh_states()` from this function...

tetienne: I'm sorry, but I have no idea what async_track_time_interval is :(

iMicknl: It is more related to Python and scoping in general. We currently have a lot of functions in our `async_setup_entry`, which is very cluttered. 

In `update_listener` I need to create a new task calling the `handle_refresh_states` function, which I can not seem to access there. I solved it with some duplicate code, but I would love to have this all a bit more clean. 
(at least it works now).

custom_components/tahoma/__init__.py
```diff
@@ -191,10 +223,33 @@ async def async_unload_entry(hass: HomeAssistant, entry: ConfigEntry):
 
 async def update_listener(hass: HomeAssistant, entry: ConfigEntry):
     """Update when config_entry options update."""
+
+    async def handle_refresh_states(*_):
```
iMicknl, line 227: This is the duplicate code I don't like that much.

iMicknl:
I am not sure if we should allow them to set this to 5 seconds, since it is a very 'expensive' call and it looks like you get 'TooManyRequests' exception when polling on a high interval. Currently we don't catch TooManyRequests exception that well in DataUpdateCoordinator.

iMicknl:
PR has been tested and is fully functional, however I am still in doubt if we should disable this task by default. 

Maybe we should add a checkbox to enable this task.

adriweb:
That looks like something that Somfy wouldn't like, especially if set with low values. I think it would be better not enabling that, especially by default, if it can get users/software in trouble.
But I suppose you should contact Somfy to make sure.

vlebourl:
Agreed, I would leave that unchecked by default and let users be responsible for making an automation to use it.

iMicknl:
> That looks like something that Somfy wouldn't like, especially if set with low values. I think it would be better not enabling that, especially by default, if it can get users/software in trouble.
> But I suppose you should contact Somfy to make sure.

Indeed, I am on the same page. Currently the widely used `tahoma-homebridge` is polling this endpoint every 30 minutes and they have many users. This endpoint is also called by tahomalink.com every time you login, however if we don't need it, we should not enable it by default.

https://github.com/dubocr/homebridge-tahoma/blob/d1a4ac3fa26adfcb06764db377f069c7bbd922ce/overkiz-api.js#L117

francodutch:
as some of you know I would love for this to be pulled in to the master - my gate closes automatically two minutes after opening, which is not caught by Tahome unless I use an automation that calls the refresh service two minutes after my gate's state has changed from closed.

vlebourl:
I think this one can be closed. This isn't something that will be added to core IMO.

PR #273, opened 2020-09-20
Change CookieJar to use defaults (unsafe=False)

iMicknl:


custom_components/tahoma/__init__.py
```diff
@@ -80,9 +80,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
     username = entry.data.get(CONF_USERNAME)
     password = entry.data.get(CONF_PASSWORD)
 
-    session = aiohttp_client.async_create_clientsession(
-        hass, cookie_jar=CookieJar(unsafe=True)
-    )
+    session = aiohttp_client.async_create_clientsession(hass, cookie_jar=CookieJar())
```
tetienne, line 83: I still think you can just do: 
`session = aiohttp_client.async_create_clientsession(hass)`

See https://docs.aiohttp.org/en/stable/client_reference.html and https://github.com/home-assistant/core/blob/53545c984bd4df54187c1d01f7af34223e8931d2/homeassistant/helpers/aiohttp_client.py

iMicknl: I will give it a try, however I remember that this didn't persist the cookie when I tried in the past :(

iMicknl: It worked this time without the CookieJar indeed.

tetienne (approved):

PR #276, opened 2020-09-27
Only remove update listener when unload was succesfull

iMicknl:


custom_components/tahoma/__init__.py
```diff
@@ -181,6 +179,7 @@ async def async_unload_entry(hass: HomeAssistant, entry: ConfigEntry):
     )
 
     if unload_ok:
+        hass.data[DOMAIN][entry.entry_id]["update_listener"]()
```
tetienne, line 182: What's the point already to call this method?

iMicknl: We register an 'update listener' with our callback function, to get notified on changes. We need to unregister this callback when the component is unloaded.

tetienne: I mean, why calling it, will unregister it?

iMicknl: Yes! Sorry, I misinterpreted your question. Have a look at the following function definition.

```
    def add_update_listener(self, listener: UpdateListenerType) -> CALLBACK_TYPE:
        """Listen for when entry is updated.

        Returns function to unlisten.
        """
```

tetienne (approved):

Issue #277, opened 2020-09-28
Handle ServerDisconnectedError during startup


laszlojakab:
- [x] **I have read the [Readme](https://github.com/iMicknl/ha-tahoma/blob/master/README.md), including the [Advanced](https://github.com/iMicknl/ha-tahoma/blob/master/README.md#advanced) section regarding debugging.**

**Is your feature request related to a problem? Please describe.**
I often restart HA for development purposes. A lot of times I get ServerDisconnectedError from tahoma integration during startup. All my covers are unavailable and I never get them back until I restart HA again (and not getting ServerDisconnectedError again). 

Stacktrace:

```
2020-09-28 21:54:37 ERROR (MainThread) [homeassistant.config_entries] Error setting up entry *****@*****.com for tahoma
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/config_entries.py", line 228, in async_setup
    result = await component.async_setup_entry(hass, self)  # type: ignore
  File "/config/custom_components/tahoma/__init__.py", line 108, in async_setup_entry
    devices=await client.get_devices(),
  File "/usr/local/lib/python3.8/site-packages/pyhoma/client.py", line 98, in get_devices
    response = await self.__get("setup/devices")
  File "/usr/local/lib/python3.8/site-packages/pyhoma/client.py", line 239, in __get
    async with self.session.get(f"{self.api_url}{endpoint}") as response:
  File "/usr/local/lib/python3.8/site-packages/aiohttp/client.py", line 1012, in __aenter__
    self._resp = await self._coro
  File "/usr/local/lib/python3.8/site-packages/aiohttp/client.py", line 504, in _request
    await resp.start(conn)
  File "/usr/local/lib/python3.8/site-packages/aiohttp/client_reqrep.py", line 847, in start
    message, payload = await self._protocol.read()  # type: ignore  # noqa
  File "/usr/local/lib/python3.8/site-packages/aiohttp/streams.py", line 591, in read
    await self._waiter
aiohttp.client_exceptions.ServerDisconnectedError
```

**Describe the solution you'd like**
It would be great to store previously discovered devices, and register them in HA in "unavailable" state during startup if server is unavailable. When connection is restored than just change the actual state of devices.

iMicknl:
Thanks for reporting. Do you have (public) internet access when you restart Home Assistant?

[...]

PR #280, opened 2020-09-29
Improve error handling on startup

iMicknl:
We now raise ConfigEntryNotReady on some exceptions, to make it retry. Not 100% sure if this will work for `too_many_requests` as well.

```
2020-09-29 23:59:35 ERROR (MainThread) [custom_components.tahoma] too_many_requests
2020-09-29 23:59:35 WARNING (MainThread) [homeassistant.config_entries] Config entry for tahoma not ready yet. Retrying in 20 seconds
2020-09-29 23:59:56 ERROR (MainThread) [custom_components.tahoma] too_many_requests
2020-09-29 23:59:56 WARNING (MainThread) [homeassistant.config_entries] Config entry for tahoma not ready yet. Retrying in 40 seconds
2020-09-30 00:00:36 ERROR (MainThread) [custom_components.tahoma] too_many_requests
2020-09-30 00:00:36 WARNING (MainThread) [homeassistant.config_entries] Config entry for tahoma not ready yet. Retrying in 80 seconds
```

### DIFF
--- a/custom_components/tahoma/__init__.py
+++ b/custom_components/tahoma/__init__.py
@@ -86,14 +86,6 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
         await client.login()
     except BadCredentialsException:
         _LOGGER.error("invalid_auth")
-        hass.async_create_task(
-            hass.config_entries.flow.async_init(
-                # Change to SOURCE_REAUTH after 116.0 release
-                DOMAIN,
-                context={"source": "reauth"},
-                data=entry.data,
-            )
-        )
         return False
     except TooManyRequestsException as exception:
         _LOGGER.error("too_many_requests")
--- a/custom_components/tahoma/__init__.py
+++ b/custom_components/tahoma/__init__.py
@@ -11,9 +11,8 @@ from pyhoma.exceptions import BadCredentialsException, TooManyRequestsException
 from pyhoma.models import Command
 import voluptuous as vol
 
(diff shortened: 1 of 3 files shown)


custom_components/tahoma/__init__.py
```diff
@@ -84,12 +84,17 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
 
     try:
         await client.login()
-    except TooManyRequestsException:
-        _LOGGER.error("too_many_requests")
-        return False
+        devices = await client.get_devices()
+        scenarios = await client.get_scenarios()
```
tetienne, line 88: Why create 2 variables? We only use them once.

iMicknl: I want to have those two API calls in the `try, catch {}` as well, however I don't want to duplicate the exception logic 3 times. This was the only solution I could think of.

tetienne: But you don't use them in your except blocks. Or perhaps you suggest that these exceptions can happen when we retrieve the devices and scenarios?

iMicknl: Yes, these exceptions can happen on retrieval of the devices and scenarios. They are needed later on in the init, however this won't be reached in case of any errors.

See https://github.com/iMicknl/ha-tahoma/issues/277.

tetienne (approved):

PR #287, opened 2020-10-05
Fix pylint warnings

iMicknl:
While working on the core PR, I found some small code issues highlighted by pylint.

There are still some style issues that are removed by our precommit hooks, however that will be fixed automatically when doing the core PR. At some point it would be good to get 100% parity with core CI/CD in this repo.

### DIFF
--- a/custom_components/tahoma/config_flow.py
+++ b/custom_components/tahoma/config_flow.py
@@ -1,5 +1,4 @@
 """Config flow for TaHoma integration."""
-from asyncio import TimeoutError
 import logging
 
 from aiohttp import ClientError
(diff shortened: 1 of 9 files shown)


custom_components/tahoma/config_flow.py
```diff
@@ -12,12 +11,8 @@
...
 from homeassistant.helpers import config_validation as cv
 
-from .const import (
-    CONF_UPDATE_INTERVAL,
-    DEFAULT_UPDATE_INTERVAL,
-    DOMAIN,
-    MIN_UPDATE_INTERVAL,
-)
+from .const import CONF_UPDATE_INTERVAL, DEFAULT_UPDATE_INTERVAL, MIN_UPDATE_INTERVAL
+from .const import DOMAIN  # pylint: disable=unused-import
```
tetienne, line 15: Pylint don't like this?

tetienne (approved):

custom_components/tahoma/config_flow.py
iMicknl: Nope... I am not 100% sure why, but every other component is disabling this warning for this line indeed.

PR #288, opened 2020-10-05
Use round() instead of int()

iMicknl:
As mentioned by the pull request in core, on DimmableLight PR.


tetienne:
I'm ok with your changes, but can you explain why?

tetienne (approved):

PR #293, opened 2020-10-08
Reverse open/close for all Awning devices

### LINKED ISSUE #292: Cover position is reversed for AwningValance
- [x] **I have read the [Readme](https://github.com/iMicknl/ha-tahoma/blob/master/README.md), including the [Advanced](https://github.com/iMicknl/ha-tahoma/blob/master/README.md#advanced) section regarding debugging.**

If cover 'volant' of my  awning is is closed the entity card shows that as open and the other way around.
This behavior is different than for other covers like 'Raffstores'

If the volant is closed display the status as 'closed' and the other way around

Thanks!
tahoma_volant
HA_volant

vlebourl:
Fix #292

### DIFF
--- a/custom_components/tahoma/cover.py
+++ b/custom_components/tahoma/cover.py
@@ -184,7 +184,7 @@ class TahomaCover(TahomaDevice, CoverEntity):
         position = 100 - kwargs.get(ATTR_POSITION, 0)
 
         # HorizontalAwning devices need a reversed position that can not be obtained via the API
-        if "Horizontal" in self.device.widget:
+        if "Horizontal" in self.device.widget or "AwningValance" == self.device.widget:
             position = kwargs.get(ATTR_POSITION, 0)
 
         await self.async_execute_command(
@@ -196,7 +196,7 @@ class TahomaCover(TahomaDevice, CoverEntity):
         position = 100 - kwargs.get(ATTR_POSITION, 0)
 
         # HorizontalAwning devices need a reversed position that can not be obtained via the API
-        if "Horizontal" in self.device.widget:
+        if "Horizontal" in self.device.widget or "AwningValance" == self.device.widget:
             position = kwargs.get(ATTR_POSITION, 0)
 
         await self.async_execute_command(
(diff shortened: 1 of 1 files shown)


tetienne (approved):

custom_components/tahoma/cover.py
```diff
@@ -157,7 +157,10 @@ def current_cover_position(self):
         if position is None or position < 0 or position > 100:
             return None
 
-        if "Horizontal" not in self.device.widget:
+        if (
+            "Horizontal" not in self.device.widget
+            and "AwningValance" != self.device.widget
```
iMicknl, line 162: This should be added to all mentions of "Horizontal" in `cover.py`, if I am not mistaken.

iMicknl (changes requested):
I would like to wait with merging this one, until we figured out if this issue is for all Horizontal Awnings after the latest update.

custom_components/tahoma/cover.py
vlebourl: done

custom_components/tahoma/cover.py
```diff
@@ -169,14 +172,19 @@ def current_cover_tilt_position(self):
...
+        else:
+            return 100 - position
 
     async def async_set_cover_position(self, **kwargs):
         """Move the cover to a specific position."""
         position = 100 - kwargs.get(ATTR_POSITION, 0)
 
         # HorizontalAwning devices need a reversed position that can not be obtained via the API
-        if "Horizontal" in self.device.widget:
+        if "Horizontal" in self.device.widget or "AwningValance" == self.device.widget:
```
iMicknl, line 187: By the way, all of those are Awnings. What if we just say:
`if self.device.uiclass == "Awning` 

It looks like they made all the Awnings reversed since the last update, which means we can maybe remove the ugly "Horizontal" hack?

vlebourl: agreed it's better

iMicknl:
I believe this is a better fix indeed! Looking at the old TaHoma component, only the `io:HorizontalAwningIOComponent` needed this reverse before the last firmware update.

`io:AwningValanceIOComponent` wasn't reversed before, but should be now. Hopefully this applies to all awnings, should we first keep this in testing or should we release and see if we break functionality for some users?

vlebourl:
I'd say merge and see. Keeping it in testing in another branch is too complex for a lot of users, we won't have enough feedback anyways

iMicknl (approved):

tetienne (approved):

tetienne:
Can we merge this PR?

iMicknl (approved):

PR #301, opened 2020-10-23
Log available commands in case of `No such command` Exception in `execute_command` service

vlebourl:
This requires https://github.com/iMicknl/python-tahoma-api/pull/51


iMicknl:
Do we really need this in our Home Assistant integration? We can easily pull the details for a user based on their details from the debug log, is there a need for them to do this themselves?

vlebourl:
My use case was to find the exact command tu use to refresh the thermostat state by using the `execute_command` service. 
As a power user, I was able to get it from the api itself, but I would have liked to have a solution to dump the commands to the log.

custom_components/tahoma/__init__.py
```diff
@@ -144,11 +145,25 @@ async def handle_execute_command(call):
...
-            "Home Assistant Service",
-        )
+        try:
+            await tahoma_coordinator.client.execute_command(
+                entity.unique_id,
+                Command(call.data.get("command"), call.data.get("args")),
+                "Home Assistant Service",
+            )
+        except Exception as e:
+            if e.args[0].startswith("No such command"):
```
tetienne, line 155: Would be nice to have a dedicated exception instead of parsing the message.

vlebourl: I know, but the client only throws Exception there

vlebourl: https://github.com/iMicknl/python-tahoma-api/blob/282ab59d56172d799ffeebb89fbfd07c7dd8fe57/pyhoma/client.py#L334

vlebourl: https://github.com/iMicknl/python-tahoma-api/pull/54

Issue #300, opened 2020-10-23
Avoid polling for stateless (RTS) devices


larsvinc:
- [x] **I have read the [Readme](https://github.com/iMicknl/ha-tahoma/blob/master/README.md), including the [Advanced](https://github.com/iMicknl/ha-tahoma/blob/master/README.md#advanced) section regarding debugging.**

**Is your feature request related to a problem? Please describe.**
It is related to https://github.com/iMicknl/ha-tahoma/issues/289

**Describe the solution you'd like**
Since all my devices are stateless, it is my understanding from the issue tracker above that the API usage with Somfy can be greatly simplified, and therefore relieve the traffic on their API. When this is the case, I believe that this is a good solution for all (simplified implementation?, reduced traffic for all).

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

tetienne:
Nice idea. We can try to detect if all the devices are stateless and/or add an option to avoid polling!

PoVPoV29:
Can you use that : controllable_name: 'rts:RollerShutterRTSComponent'
RTS is stateless : radio signal in one way
IO is not stateless : radio signal in two way

tetienne:
@PoVPoV29 We don’t have to check the protocol, Somfy returns a states object. We just have to check its presence or not.

tetienne:
@PoVPoV29 @larsvinc If you have RTS devices, can I ask you to test this branch and in particular the stop command?

larsvinc:
Hi,

Tested the branch. No devices were found.

PR #302, opened 2020-10-23
Don’t poll if stateless #300

tetienne:


iMicknl:
@tetienne nice fix! My only concern is that the event listener is still registered and still polled every second after an execution, thus this would not fully fix the #300 yet.

However, we could merge this already as a good start.

tetienne:
@iMicknl Indeed, I can add a condition [against the state](https://github.com/iMicknl/ha-tahoma/blob/d81fc1b238272f4e6088da57c9cc3c53a5afb52d/custom_components/tahoma/tahoma_device.py#L67) when we send a [command ](https://github.com/iMicknl/ha-tahoma/blob/master/custom_components/tahoma/tahoma_device.py#L179)to avoid a call to the coordinator just after.

custom_components/tahoma/__init__.py
```diff
@@ -197,3 +203,8 @@ async def update_listener(hass: HomeAssistant, entry: ConfigEntry):
         coordinator.original_update_interval = new_update_interval
 
         await coordinator.async_refresh()
+
+
+async def is_stateless(devices: List[Device]) -> bool:
+    """Return true if all the devices are stateless."""
+    return all(device.states is None or len(device.states) == 0 for device in devices)
```
iMicknl, line 210: We could check here for devices with assumed_state as well? What is the benefit of checking for states again?

custom_components/tahoma/tahoma_device.py
```diff
@@ -170,6 +170,9 @@ async def async_execute_command(self, command_name: str, *args: Any):
             _LOGGER.error(exception)
             return
 
+        if self.assumed_state:
```
iMicknl, line 173: I only need to see if this doesn't break the stop functionality. Not sure if the executions are currently cancelled to stop RTS devices.

custom_components/tahoma/__init__.py
tetienne: We haven’t yet build the tahoma devices ;)

custom_components/tahoma/tahoma_device.py
tetienne: Good point. @vlebourl You have RTS covers right?

iMicknl: @tetienne and maybe we should also add logic in `is_opening` and `is_closing` in cover.py to only do this for IO devices! (another PR)

tetienne: I don’t think so. The code is already protected as we check the execution list. Except if you are concerned about performance?

iMicknl: @tetienne I was indeed wondering about performance, but I thought it was doing an API call. It is just retrieving the values from the local store, which is fine to keep like this.

custom_components/tahoma/__init__.py
```diff
@@ -98,7 +99,12 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
         _LOGGER.exception(exception)
         return False
 
-    update_interval = entry.options.get(CONF_UPDATE_INTERVAL, DEFAULT_UPDATE_INTERVAL)
+    if is_stateless(devices):
+        update_interval = None
+    else:
+        update_interval = entry.options.get(
+            CONF_UPDATE_INTERVAL, DEFAULT_UPDATE_INTERVAL
+        )
```
tetienne, line 107: I don’t like this, I will refactor it.

custom_components/tahoma/tahoma_device.py
tetienne: So now we have only to know if it breaks the stop feature for RTS devices right?

tetienne: @iMicknl A possible solution would be to store the exec_id directly into the device as a variable. 

```Python
async def async_execute_command(self, command_name: str, *args: Any):
        """Execute device command in async context."""
        try:
            self.last_exec_id= await self.coordinator.client.execute_command(
                self.device.deviceurl,
                Command(command_name, list(args)),
                "Home Assistant",
            )
        except Exception as exception:  # pylint: disable=broad-except
            _LOGGER.error(exception)
            return

        if self.assumed_state:
            return

        # ExecutionRegisteredEvent doesn't contain the deviceurl, thus we need to register it here
        self.coordinator.executions[self.last_exec_id] = {
            "deviceurl": self.device.deviceurl,
            "command_name": command_name,
        }

        await self.coordinator.async_refresh()
```

vlebourl (dismissed):

vlebourl (approved):

iMicknl:
Thanks! I have used this as a base for the core pr: https://github.com/home-assistant/core/pull/49639.

PR #303, opened 2020-10-23
Run CI/CD job only on pull request

tetienne:
@iMicknl If you forbid direct push on master, we can ease the workflow and avoid to run twice the checks (one for the PR and one for the push).
I also removed the scheduled job, their is no need to do this except for dynamic check like hassfest or hacs.


iMicknl:
Good catch! Should we not make sure this one is run on master as well? (Is that possible?).
In case we merge an (outdated) PR, which causes conflicts in master branch.

iMicknl:
By the way, doesn't push mean that it will be run on every commit in a PR? Or is this still the same when triggered by `pull_request`.

vlebourl (approved):

tetienne:
@iMicknl If you forbid direct push on the master branch, it will be ok. Triggering only on ̏pull_request` will analye every commit on a PR (it’s why you see dup check).

iMicknl (approved):

PR #309, opened 2020-10-26
Update pre-commit version

tetienne:


iMicknl:
Good one! I faced this indeed when making the core PR...

iMicknl (dismissed):

tetienne:
It would be nice if HACS give a way to keep in sync the linters between custom component and the Core using an action for instance.

iMicknl (approved):

PR #316, opened 2020-12-03
Implement Coordinator entity

tetienne:


iMicknl (approved):
Looks great, can be merged if it has been tested :D

tetienne:
@iMicknl Yes, tested locally. The methods I removed are exactly the same that the ones you can find within CoordinatorEntity.

PR #318, opened 2020-12-04
Print HomeKit code in log while in debug mode

iMicknl:
@vlebourl @tetienne first version, are you ok to merge it like this for the logging, or should I rewrite it to make it nicer?


tetienne:
Can you please move the code within a dedicated method?

custom_components/tahoma/__init__.py
```diff
@@ -143,6 +146,9 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
                 device.widget,
             )
 
+        if device.widget == HOMEKIT_STACK:
```
tetienne, line 149: I would prefer to have the condition inside the method too.

iMicknl: Sorry for merging in advance @tetienne, I will work on it in a future PR...

PR #327, opened 2020-12-18
Support 'suddenDropMode' state for Somfy Thermostat

vlebourl:


tetienne:
OK for me if you tested it on you own thermostat.

tetienne (approved):

iMicknl (approved):
Can multiple modes be mapped to preset none? I am not sure if I understand this change, but ok to merge if you tested it :D

vlebourl:
I'm not sure how it works honestly. sudden drop is not something one can set on its own, but automatically detected by the thermostat. In that case, the heating is paused and the mode set to "sudden drop", which wasn't existing and thus caused an error in ha. this is just a quick fix, but I haven't been able to trigger that mode again...

iMicknl:
@vlebourl can this one be merged?

PR #329, opened 2020-12-18
Use existing web session

tetienne:
We have to use async_get_clientsession instead of async_create_clientsession. The last one creates a session, the former uses an existing one.

See also:
* https://github.com/home-assistant/core/search?q=async_create_clientsession
* https://github.com/home-assistant/core/search?q=async_get_clientsession

### DIFF
--- a/custom_components/tahoma/__init__.py
+++ b/custom_components/tahoma/__init__.py
@@ -10,7 +10,8 @@ from homeassistant.config_entries import SOURCE_IMPORT, ConfigEntry
 from homeassistant.const import CONF_EXCLUDE, CONF_PASSWORD, CONF_USERNAME
 from homeassistant.core import HomeAssistant
 from homeassistant.exceptions import ConfigEntryNotReady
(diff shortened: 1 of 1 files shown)


vlebourl (approved):

custom_components/tahoma/__init__.py
```diff
@@ -86,7 +87,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
     username = entry.data.get(CONF_USERNAME)
     password = entry.data.get(CONF_PASSWORD)
 
-    session = aiohttp_client.async_create_clientsession(hass)
+    session = async_get_clientsession(hass)
```
vlebourl, line 90: don't we need to create a session at some point before doing a get? 
if not, feel free to merge!

vlebourl (approved):

custom_components/tahoma/__init__.py
tetienne: Nope HA does it for us. That's why it's nice! I tested locally to be sure.

PR #332, opened 2021-01-12
Support Cozytouch, Rexel and Connexoon RTS.

iMicknl:
This PR adds support for other hubs, based on the same Overkiz API.


custom_components/tahoma/__init__.py
```diff
@@ -87,8 +89,18 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
     username = entry.data.get(CONF_USERNAME)
     password = entry.data.get(CONF_PASSWORD)
 
+    if entry.data.get(CONF_HUB):
+        endpoint = SUPPORTED_ENDPOINTS[entry.data.get(CONF_HUB)]
+    else:
+        endpoint = None
```
tetienne, line 95: Can we fallback to the classic endpoint instead? It will ease to understand the code.

iMicknl: @tetienne good one, I will simplify this. When passing 'None', it will default to the tahomalink.com endpoint since it is default behaviour of pyhoma. However this is a bit like a black box in this way indeed.

iMicknl: @tetienne I believe I solved in a better way now.

custom_components/tahoma/config_flow.py
```diff
@@ -41,9 +52,15 @@ async def async_validate_input(self, user_input):
         username = user_input.get(CONF_USERNAME)
         password = user_input.get(CONF_PASSWORD)
 
-        async with TahomaClient(username, password) as client:
+        hub = user_input.get(CONF_HUB) or DEFAULT_HUB
+        endpoint = SUPPORTED_ENDPOINTS[hub]
+
+        async with TahomaClient(username, password, api_url=endpoint) as client:
```
vlebourl, line 58: If we support multiple endpoints, shouldn't we rename this (and the whole integration???) as it's not restricted to TaHoma anymore?

vlebourl (approved):

custom_components/tahoma/config_flow.py
iMicknl: I have been thinking about rebranding to Overkiz (the platform used by all companies), however for the sake of discoverability I would prefer to keep it Somfy TaHoma.

We can add this context to the docs and I will work on auto discover so that it will be found more easily when added to core.

custom_components/tahoma/const.py
```diff
@@ -8,8 +8,18 @@
...
+CONF_HUB = "hub"
 DOMAIN = "tahoma"
 
+DEFAULT_HUB = "Somfy TaHoma"
+SUPPORTED_ENDPOINTS = {
+    "Cozytouch": "https://ha110-1.overkiz.com/enduser-mobile-web/enduserAPI/",
+    "Somfy TaHoma": "https://tahomalink.com/enduser-mobile-web/enduserAPI/",
+    "Somfy Connexoon IO": "https://tahomalink.com/enduser-mobile-web/enduserAPI/",
+    "Somfy Connexoon RTS": "https://ha201-1.overkiz.com/enduser-mobile-web/enduserAPI/",
+    "Rexel": "https://ha112-1.overkiz.com/enduser-mobile-web/enduserAPI/",
```
tetienne, line 20: Would be nice later to add this within pyhoma.

tetienne (approved):

custom_components/tahoma/const.py
iMicknl: @tetienne good one! Would be a good improvement in the future.

PR #333, opened 2021-01-12
Add gateways to Device Registry

iMicknl:
First try to add gateways to the device registry as well. I saw you did do this for the Somfy integration, @tetienne. Any specific reason for it? (or just cosmetic).

Screen Shot 2021-01-12 at 20 58 07

Screen Shot 2021-01-12 at 21 00 35


tetienne:
Yes it was just cosmetic. For the moment it's just an useless device 😅
Perhaps later, Core team will use it for something.

tetienne (commented):
You have to add a via_hub attribute to the device info too.

iMicknl:
> You have to add a via_hub attribute to the device info too.

Thanks, I was just looking into this. But I can not seem to find a relation between the hub and the devices via the TaHoma API...


Another thought (for a new PR) is to add sensors to the hub, for example reflecting `updateStatus / upToDate` and `connectivity -> status`.

tetienne:
Perhaps because there is no need for them to return a parent id. We can assume that for each device in the response the parent is the only hub available.

vlebourl:
is it really necessary? isn't it cluttering the UI with useless info?

iMicknl:
> is it really necessary? isn't it cluttering the UI with useless info?

You won't see it much in the UI, but it will allow us to also detect if we already have the integration configured for autodiscover. Autodiscover is able to receive the ID of the gateway, via the hostname.

vlebourl (approved):

iMicknl:
> You have to add a via_hub attribute to the device info too.
Done! However, I don't see any mention of it in the interface, is this property still required?

iMicknl:
> > You have to add a via_hub attribute to the device info too.
> 
> Done! However, I don't see any mention of it in the interface, is this property still required?

Apparently it is `via_device`, instead of `via_hub`. @tetienne, you have something to change for Somfy as well :D
https://developers.home-assistant.io/docs/device_registry_index/#what-is-a-device

tetienne:
Seriously? Why nobody told me that's was wrong?!

iMicknl:
Screen Shot 2021-01-15 at 17 06 13

Screen Shot 2021-01-15 at 17 05 58

Tested with TaHoma hub and Connexoon Australia! This PR is ready to be merged I believe :)

custom_components/tahoma/manifest.json
```diff
@@ -4,7 +4,7 @@
   "config_flow": true,
   "documentation": "https://www.home-assistant.io/integrations/tahoma",
   "requirements": [
-    "pyhoma==0.5.1"
+    "pyhoma==0.5.2"
```
vlebourl, line 7: shouldn't this be 0.5.3?

requirements.test.txt
```diff
@@ -7,4 +7,4 @@ pytest-cov<3.0.0
 pytest-homeassistant
 
 # from our manifest.json for our Custom Component
-pyhoma==0.5.1
\ No newline at end of file
+pyhoma==0.5.2
```
vlebourl, line 10: same here?

vlebourl (approved):

custom_components/tahoma/manifest.json
iMicknl: Done.

iMicknl:
@Pol2Tls, @browetd; could you give this branch a try? I would like to understand if it works with Cozytouch as well.

https://github.com/iMicknl/ha-tahoma/archive/feature/add_gateway.zip

iMicknl:
> @iMicknl I tried again this morning but still 12 equipements and 31 entities... No climate entity... But the bridge is not on the same Lan than my raspberry pi (it will be in a couple of months), I do not know if this is important or not for the test...

Thanks @browetd. This change was indeed not for the climate entity, hopefully we will be able to spend some time on that soon. Thanks for testing :).


Other news;
```
2021-01-15 17:50:54 ERROR (MainThread) [homeassistant.config_entries] Error setting up entry xxxxxx@gmail.com for tahoma
  File "/usr/src/homeassistant/homeassistant/config_entries.py", line 236, in async_setup
    result = await component.async_setup_entry(hass, self)  # type: ignore
  File "/config/custom_components/tahoma/__init__.py", line 206, in async_setup_entry
    beautify_name(gateway.sub_type.name) if gateway.sub_type else None
AttributeError: sub_type
```

For Cozytouch (without sub_type) I need to make a change in python-tahoma-api first, before merge.

custom_components/tahoma/__init__.py
```diff
@@ -199,6 +205,36 @@ async def handle_execute_command(call):
...
+        _LOGGER.debug(
+            "Gateway detected (%s - %s - %s)",
+            gateway.id,
+            gateway.type,
+            gateway.sub_type,
+        )
+
+        gateway_model = (
+            beautify_name(gateway.sub_type.name)
+            if isinstance(gateway.sub_type, Enum)
```
tetienne, line 220: It can be something else than an Enum?

iMicknl: Unfortunately it can be something else... (https://github.com/iMicknl/python-tahoma-api/blob/78587f4f9547b2621a1123817efb2f2a66f21d9b/pyhoma/models.py#L394-L402).

However, maybe we should address it there. And throw a None + warning that we don't have the value in Enum.

tetienne: In Pyhoma the code looks good. But here it sounds weird :S

iMicknl: I can also do this in `beautify_name()`, but I would prefer to just pass a string to that to keep is reusable.

PR #342, opened 2021-01-14
Make 'execute_command' service an admin service

iMicknl:
Execute Command is a very powerful service and should only be accessible to admins.


vlebourl (approved):
if tested and confirmed working, fine by me

iMicknl:
> if tested and confirmed working, fine by me

Will double check before merge 👍🏻. Should have been tested a while ago.

PR #328, opened 2020-12-18
Add battery sensor as seperate entity

vlebourl:
image

PR #343, opened 2021-01-14
Group entities by physical device

iMicknl:
Functionality from #328 by @vlebourl


iMicknl:
@vlebourl could you give this a try? I don't have any devices that have multiple entities...

custom_components/tahoma/tahoma_device.py
```diff
@@ -66,11 +66,7 @@ def assumed_state(self) -> bool:
     @property
     def device_state_attributes(self) -> Dict[str, Any]:
         """Return the state attributes of the device."""
-        attr = {
-            "ui_class": self.device.ui_class,
```
iMicknl, line 70: Those details are added as device information and in the debug log. Since last versions the attributes are added to the front-end, thus this is cluttering the view.

custom_components/tahoma/sensor.py
```diff
@@ -26,12 +27,6 @@
 from .const import DOMAIN
 from .tahoma_device import TahomaDevice
 
-try:  # TODO: Remove for core PR. This ensures compatibility with <0.115
```
iMicknl, line 29: We use other code that is only compatible with the latest core versions.

custom_components/tahoma/__init__.py
```diff
@@ -147,7 +147,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
         if platform:
             entities[platform].append(device)
             _LOGGER.debug(
-                "Added TaHoma device (%s - %s - %s - %s)",
+                "Added device (%s - %s - %s - %s)",
```
iMicknl, line 150: TaHoma is already in the log, before the message. This applies to all changes here.

vlebourl:
> @vlebourl could you give this a try? I don't have any devices that have multiple entities...

I'll try that tomorrow! But looks good to me.

vlebourl:
that works for me!
image

vlebourl (dismissed):

iMicknl:
@vlebourl I see that I need to copy the device info as well, looks like it is now displaying the last one.

vlebourl:
> @vlebourl I see that I need to copy the device info as well, looks like it is now displaying the last one.

ohyeah... missed that one

vlebourl:
> > @vlebourl I see that I need to copy the device info as well, looks like it is now displaying the last one.
> 
> ohyeah... missed that one

That's why somewhere I retrieve the `base_url#1` label to get the device info.

iMicknl:
> > > @vlebourl I see that I need to copy the device info as well, looks like it is now displaying the last one.
> > 
> > 
> > ohyeah... missed that one
> 
> That's why somewhere I retrieve the `base_url#1` label to get the device info.

Did this work in your previous PR?

vlebourl:
hmm scratch that, that's actually the `model` part in the return of `tahoma_device` `device_info`

iMicknl:
@vlebourl we can pull model and firmware from our parent device (#1) as well. I am not able to test it, but what if we just leave it empty for devices not being #1? Could you try that if I add it to this pr?

vlebourl:
only `#1` info shows up now: 
image
that's fine IMO

vlebourl:
but firmware is wrong...

iMicknl:
@vlebourl could you perhaps give the latest version a try? If this doesn't work, I am not sure how to solve it :-). It is hard to fix without having the device present.

iMicknl:
While removing old branches, I came across https://github.com/iMicknl/ha-tahoma/pull/152. Was this simple approach not good enough? :D

custom_components/tahoma/tahoma_device.py
```diff
@@ -109,14 +105,45 @@ def device_info(self) -> Dict[str, Any]:
...
+        device_url = self.device_url
+
+        # Some devices, such as the Smart Thermostat have several devices in one physical device,
+        # with same device url, terminated by '#' and a number.
+        # We use the base device url as the device identifier.
+        if "#" in self.device_url:
+            device_url, _ = self.device.deviceurl.split("#")
+
+            if not self.device_url.endswith("#1"):
+                entity_registry = self.hass.data["entity_registry"]
```
iMicknl, line 118:
```suggestion
                entity_registry = await self.hass.helpers.entity_registry.async_get_registry()
```

tetienne:
@iMicknl If I remember well, it was because the name of the device was not correct. But indeed, the amount of modification between this PR and mine are not the same at all :)
@lebourl how dit you solve this: https://github.com/iMicknl/ha-tahoma/pull/152/files#r455789429 ?

iMicknl:
> @iMicknl If I remember well, it was because the name of the device was not correct. But indeed, the amount of modification between this PR and mine are not the same at all :)

Ahh, that is indeed solved in this PR. The amount of code is indeed more, but it is not that much. Most of the code is around retrieving the entity id / name of the original device.

custom_components/tahoma/tahoma_device.py
```diff
@@ -109,14 +105,45 @@ def device_info(self) -> Dict[str, Any]:
...
+        device_info.update(
+            {
+                "name": name,
+                "manufacturer": manufacturer,
+                "model": model,
+                "sw_version": self.device.controllable_name,
+            }
+        )
+
+        return device_info
```
tetienne, line 145: Can you try to split the logic in several methods? That's not easy to understand at the first look.

custom_components/tahoma/tahoma_device.py
```diff
@@ -109,14 +105,45 @@ def device_info(self) -> Dict[str, Any]:
...
+                    None,
+                )
+
+        device_info = {
+            "identifiers": {(DOMAIN, device_url)},
         }
 
+        # Only return device identifier for sub devices.
+        if "#" in self.device_url and not self.device_url.endswith("#1"):
+            return device_info
```
tetienne, line 134: Why not return the other information? It introduces weird behavior?

custom_components/tahoma/tahoma_device.py
```diff
@@ -109,14 +105,45 @@ def device_info(self) -> Dict[str, Any]:
...
         model = self.select_state(CORE_MODEL_STATE) or self.device.widget
 
-        return {
-            "identifiers": {(DOMAIN, self.unique_id)},
-            "manufacturer": manufacturer,
-            "name": self.name,
-            "model": model,
-            "sw_version": self.device.controllable_name,
+        name = self.name
+        device_url = self.device_url
```
tetienne, line 109: From what I see, you don't need this. You never use it. The next time we see it, you change its value [here]( https://github.com/iMicknl/ha-tahoma/pull/343/files#diff-14ad169bb8b77afd973fb894391dd4b0e629c3ebef801e129e618e09ce79a444R115) is enough.

iMicknl: We need it for devices that have no subdevices. I will see if I can make the code more clear.

custom_components/tahoma/tahoma_device.py
iMicknl: HA does use the device info from the last entity of a device, while we prefer to show information of the first device..

custom_components/tahoma/tahoma_device.py
tetienne: At least perhaps rename it to something like `main_device_url`?

custom_components/tahoma/tahoma_device.py
tetienne: The following information are not the same across the devices?
```
{
                "name": name,
                "manufacturer": manufacturer,
                "model": model,
                "sw_version": self.device.controllable_name,
            }
```

custom_components/tahoma/tahoma_device.py
iMicknl: I pushed a new version which is totally different, thanks for your review. Hopefully it still works, unfortunately I am not able to test..

custom_components/tahoma/tahoma_device.py
iMicknl: Not sure... Controllable name is different per Somfy entity for sure (io: xxx).

iMicknl:
@vlebourl could you perhaps give this PR a try? I am not able to test it and would be good to check before merge.

custom_components/tahoma/tahoma_device.py
```diff
@@ -107,13 +103,23 @@ def device_state_attributes(self) -> Dict[str, Any]:
...
+                "identifiers": {(DOMAIN, self.get_base_device_url())},
+            }
+
         manufacturer = self.select_state(CORE_MANUFACTURER_NAME_STATE) or "Somfy"
         model = self.select_state(CORE_MODEL_STATE) or self.device.widget
 
         return {
-            "identifiers": {(DOMAIN, self.unique_id)},
-            "manufacturer": manufacturer,
+            "identifiers": {(DOMAIN, self.device_url)},
```
vlebourl, line 120: Device with `#1` in the url should also return the base url to be merged with subsequent subdevices.

```suggestion
            "identifiers": {(DOMAIN, self.get_base_device_url())},
```

following, `get_base_device_url()` should be modified as well, see following comment.

custom_components/tahoma/tahoma_device.py
```diff
@@ -169,6 +175,11 @@ async def async_cancel_command(self, exec_id: str):
         """Cancel device command in async context."""
         await self.coordinator.client.cancel_command(exec_id)
 
+    def get_base_device_url(self):
+        """Return base device url."""
+        device_url, _ = self.device.deviceurl.split("#")
```
vlebourl, line 180: following previous comment:
```suggestion
        device_url, _ = (
            self.device_url.split("#")
            if "#" in self.device_url
            else (self.device_url, "")
        )
```

vlebourl (changes requested):

vlebourl:
Never mind previous comments, I propose the following solution, better IMO

custom_components/tahoma/tahoma_device.py
```diff
@@ -107,13 +103,23 @@ def device_state_attributes(self) -> Dict[str, Any]:
...
     def device_info(self) -> Dict[str, Any]:
         """Return device registry information for this entity."""
+        # Some devices, such as the Smart Thermostat have several devices in one physical device,
+        # with same device url, terminated by '#' and a number.
+        # We use the base device url as the device identifier.
+        if "#" in self.device_url and not self.device_url.endswith("#1"):
+
+            # Only return the url of the base device, to use main device name and model.
+            return {
+                "identifiers": {(DOMAIN, self.get_base_device_url())},
```
vlebourl, line 113:
```suggestion
                "identifiers": {(DOMAIN, self. device_base_url)},
```

vlebourl (changes requested):

custom_components/tahoma/tahoma_device.py
iMicknl: Thanks!

I will hold off a bit more on this PR, since it needs to take into account scenario's where deviceUrl#3 is supported, but where deviceUrl#1 is not supported (climate for example). In this case it should have some kind of a fallback.

vlebourl (approved):

PR #348, opened 2021-01-16
fix bug in hvac modes introduced in #336

vlebourl:


iMicknl (approved):
Oops!

iMicknl:
@vlebourl I was a bit confused by https://github.com/iMicknl/ha-tahoma/blob/master/custom_components/tahoma/climate_devices/atlantic_pass_apcdhw.py, they have the dict reversed...

Maybe we should just reverse the dict on one of them.

PR #388, opened 2021-02-16
Fix exception for devices who return JSON_ARRAY/JSON.OBJECT as state

vlebourl:


iMicknl:
Looks good. I wonder if this would work as well:

```python
TYPES = {
    DataType.NONE: None,
    DataType.INTEGER: int,
    DataType.DATE: int,
    DataType.STRING: str,
    DataType.FLOAT: float,
    DataType.BOOLEAN: bool,
	DataType.JSON_ARRAY: json.loads,
	DataType.JSON_OBJECT: json.loads,
}
```

vlebourl:
I don't see why not!

iMicknl (approved):
Should we ask the user to test it, before merging. :-)

PR #391, opened 2021-02-18
fix unit measure for CO2Sensor

### LINKED ISSUE #390: CO2 sensor added but no entity created
**Describe the bug**
I have a CO2 sensor available with my Atlantic Optimocosy device and the sensor is added by the component : 
```
2021-02-17 14:42:18 DEBUG (MainThread) [custom_components.tahoma] Added device (io:CO2IOSystemDeviceSensor - AirSensor - CO2Sensor - io://xxxx-xxxx-xxxx/15021786#2)
```
But I don't have any entity created with this sensor.

**To Reproduce**
I made some investigations and the problem is that the CO2Sensor doesn't have the "CORE_MEASURED_VALUE_TYPE" needed for the select of unit of measurement.
https://github.com/iMicknl/ha-tahoma/blob/3c43f1fc585c62cb7ee547bb23dc47bdcba1f55b/custom_components/tahoma/sensor.py#L135-L140

Here are the attributes of the CO2Sensor : 
```
      {
        "creationTime": 1482935460000,
        "lastUpdateTime": 1482935460000,
        "label": "IO (15021786#2)",
        "deviceURL": "io://xxxx-xxxx-xxxx/15021786#2",
        "shortcut": false,
        "controllableName": "io:CO2IOSystemDeviceSensor",
        "definition": {
          "commands": [],
          "states": [
            {
              "type": "ContinuousState",
              "qualifiedName": "core:CO2ConcentrationState"
            },
            {
              "values": [
                "available",
                "unavailable"
              ],
              "type": "DiscreteState",
              "qualifiedName": "core:StatusState"
            }
          ],
          "dataProperties": [],
          "widgetName": "CO2Sensor",
          "uiClass": "AirSensor",
          "qualifiedName": "io:CO2IOSystemDeviceSensor",
          "type": "SENSOR"
        },
        "states": [
          {
            "name": "core:StatusState",
            "type": 3,
            "value": "available"
          },
          {
            "name": "core:CO2ConcentrationState",
            "type": 1,
            "value": 601
          }
        ],
        "attributes": [
          {
            "type": 3,
            "value": "mainSupply"
          }
        ],
        "available": true,
        "enabled": true,
        "placeOID": "27c236f2-8df0-49d6-80e8-bcdcae9f9dd4",
        "widget": "CO2Sensor",
        "type": 2,
        "oid": "55288adc-5b26-495c-afeb-615f405afcf8",
        "uiClass": "AirSensor"
      },
```
**Expected behavior**
I made a try to match the unit with the Device_class and now the entity is available.
```
    def unit_of_measurement(self):
        """Return the unit of measurement of this entity, if any."""
        UNITS_CLASS = {
            DEVICE_CLASS_CO2: CONCENTRATION_PARTS_PER_MILLION,
            DEVICE_CLASS_CO: CONCENTRATION_PARTS_PER_MILLION,
        }
        if self.device_class in UNITS_CLASS:
            return UNITS_CLASS.get(self.device_class)

        if self.device.attributes:
            attribute = self.device.attributes[CORE_MEASURED_VALUE_TYPE]
            return UNITS.get(attribute.value)
        return None
```
I don't know if the COSensor have the same problem but it's the same ppm unit so it maybe usefull to add it as exception too.

Although, I find a problem with the icons : 
https://github.com/iMicknl/ha-tahoma/blob/3c43f1fc585c62cb7ee547bb23dc47bdcba1f55b/custom_components/tahoma/sensor.py#L52-L53
For COSensor, the correct one is "mdi:molecule-co", and for CO2Sensor, the correct one is "mdi:molecule-co2"

And the result : 
Capture d’écran_2021-02-17_15-22-32

tillstaff:
fix #390

### DIFF
--- a/custom_components/tahoma/sensor.py
+++ b/custom_components/tahoma/sensor.py
@@ -4,6 +4,7 @@ from typing import Optional
 
 from homeassistant.components.sensor import DOMAIN as SENSOR
 from homeassistant.const import (
+    CONCENTRATION_PARTS_PER_MILLION,
     DEVICE_CLASS_HUMIDITY,
     DEVICE_CLASS_ILLUMINANCE,
     DEVICE_CLASS_POWER,
@@ -49,8 +50,8 @@ DEVICE_CLASS_CO2 = "co2"
 DEVICE_CLASS_SUN_ENERGY = "sun_energy"
 DEVICE_CLASS_WIND_SPEED = "wind_speed"
 
-ICON_AIR_FILTER = "mdi:air-filter"
-ICON_PERIODIC_TABLE_CO2 = "mdi:periodic-table-co2"
+ICON_MOLECULE_CO = "mdi:molecule-co"
+ICON_MOLECULE_CO2 = "mdi:molecule-co2"
 ICON_SOLAR_POWER = "mdi:solar-power"
 ICON_WEATHER_WINDY = "mdi:weather-windy"
 
@@ -91,6 +92,11 @@ UNITS = {
     "core:FossilEnergyInKWh": ENERGY_KILO_WATT_HOUR,
     "core:FossilEnergyInMWh": f"M{ENERGY_WATT_HOUR}",
     "meters_seconds": SPEED_METERS_PER_SECOND,
+}
+
+UNITS_BY_DEVICE_CLASS = {
+    DEVICE_CLASS_CO2: CONCENTRATION_PARTS_PER_MILLION,
+    DEVICE_CLASS_CO: CONCENTRATION_PARTS_PER_MILLION,
 }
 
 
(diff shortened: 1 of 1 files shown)


custom_components/tahoma/sensor.py
```diff
@@ -134,6 +140,8 @@ def state(self):
     @property
     def unit_of_measurement(self):
         """Return the unit of measurement of this entity, if any."""
+        if self.device_class in UNITS_BY_DEVICE_CLASS:
```
iMicknl, line 143: I would prefer to have this after we check for `self.device.attributes`, since this should be leading. This would require a small addition to the if check below.

iMicknl (changes requested):
Thanks for your first contribution, much appreciated. Hereby a small change request, let me know if you need any help.

custom_components/tahoma/sensor.py
tillstaff: I already tried that but the check for self.device.attributes will match my CO2Sensor because the CO2Sensor has attributes but not the good one "CORE_MEASURED_VALUE_TYPE".
Maybe the "if self.devices.attributes" could be more specific with a check that at least one attribute is named "CORE_MEASURED_VALUE_TYPE"?

iMicknl: @tillstaff, sorry for the slow reply. I have been working with core to have a CO/CO2 device class added! :) 
https://github.com/home-assistant/core/pull/47487

iMicknl:
I will close this one in favour of https://github.com/iMicknl/ha-tahoma/pull/400. Thanks for your first PR!! 👍🏻

PR #416, opened 2021-04-13
Check for 'unavailable' while updating temperature sensor in Climate device

vlebourl:


custom_components/tahoma/climate_devices/somfy_thermostat.py
```diff
@@ -141,7 +142,11 @@ async def _async_temp_sensor_changed(self, entity_id, old_state, new_state) -> N
     @callback
     def update_temp(self, state):
         """Update thermostat with latest state from sensor."""
-        if state is None or state.state == STATE_UNKNOWN:
+        if (
+            state is None
+            or state.state == STATE_UNKNOWN
```
iMicknl, line 147:
```suggestion
            or state.state in [STATE_UNKNOWN, STATE_UNAVAILABLE]
```

Could we do something like this perhaps?

vlebourl: done

iMicknl:
Dismissing test (3.8) since it is not related to this PR.

PR #418, opened 2021-04-13
Refactor supported hubs/servers and load them from pyhoma

iMicknl:
Backported from core PR. Eventually it would be good to maybe rename this to 'manufacturer', since I removed the model names (Tahoma / Connexoon) from the list. Apparently Somfy has server regions, which I identified as Connexoon RTS in the past...


custom_components/tahoma/const.py
```diff
@@ -9,28 +9,29 @@
 from homeassistant.components.switch import DOMAIN as SWITCH
 from homeassistant.components.water_heater import DOMAIN as WATER_HEATER
 
-CONF_HUB = "hub"
 DOMAIN = "tahoma"
 
-DEFAULT_HUB = "Somfy TaHoma"
+CONF_HUB = "hub"
+DEFAULT_HUB = "Somfy (Europe)"
```
iMicknl, line 15: This will break behavior for current users. I still need to add a fallback in case the hub can't be found. Don't want to add a migration to this custom integration.

iMicknl:
I should add 
```
"Hi Kumo (Europe)": "https://ha117-1.overkiz.com/enduser-mobile-web/enduserAPI/",
"Hi Kumo (Asia)": "https://ha203-1.overkiz.com/enduser-mobile-web/enduserAPI/",
"Hi Kumo (Oceania)": "https://ha203-1.overkiz.com/enduser-mobile-web/enduserAPI/",
``` as well!

**Comment by tetienne** [comment]
Where did you find these ones?

File: custom_components/tahoma/const.py (hunk 1 above)
Comment (iMicknl) [reply custom_components/tahoma/const.py:15]: @tetienne @vlebourl what do you think would be the best?

I would love to rebrand this repo + the integration to overkiz. Perhaps we can do this at the same moment? It will be a breaking change, but at least gives us the opportunity to do two breaking changes at the same time ;-).

**Comment by iMicknl** [comment]
@tetienne maybe we should have a different approach here, where we change to a different dictionary model with keys. It would be good to be able to change the cosmetic name, without breaking the integration in the future. Any suggestions?

**Comment by tetienne** [comment]
> @tetienne maybe we should have a different approach here, where we change to a different dictionary model with keys. It would be good to be able to change the cosmetic name, without breaking the integration in the future. Any suggestions?

What about somethink like:
```Python
endpoints = {
  "endpoint_1": ("Tahoma", "https://..."),
  "endpoint_2: ("Overkiz", "https://...),
}
```
To the end user we display `endpoints[i][0]`, and when it selects an endpoint we look for the key matching its choice in `endpoints`.

iMicknl:
> To the end user we display `endpoints[i][0]`, and when it selects an endpoint we look for the key matching its choice in `endpoints`.

Would it make sense to have them typed? So we don't have to pull them like endpoints[i][0]? I am not sure how to type them easily, but perhaps just like this? And maybe we should just add this to python-tahoma-api finally?

```python3
@dataclass
class Hub:
    """Class to describe a hub."""
    endpoint: str
    name: str
    manufacturer: str


endpoints = {
    "atlantic_cozytouch": Hub(
        name="Atlantic Cozytouch",
        endpoint="https://ha110-1.overkiz.com/enduser-mobile-web/enduserAPI/",
        manufacturer="Atlantic",
    ),
}
```

iMicknl:
A Dutch user just verified that he has to use the Europe server with this Connexoon RTS. The new server list is correct thus :-).

custom_components/tahoma/__init__.py
```diff
@@ -90,21 +89,50 @@ async def async_setup(hass: HomeAssistant, config: dict):
     return True
 
 
+async def async_migrate_entry(hass: HomeAssistant, config_entry: ConfigEntry):
```
iMicknl, line 92: Works like a charm! Doesn't make this PR breaking anymore, and easy to remove for core PR.

```
2021-08-09 22:41:33 DEBUG (MainThread) [custom_components.tahoma] Migrating from version 1
2021-08-09 22:41:33 INFO (MainThread) [custom_components.tahoma] Migrated Somfy TaHoma to somfy_europe
2021-08-09 22:41:33 INFO (MainThread) [custom_components.tahoma] Migration to version 2 successful
```

tetienne: Nice!

tetienne (approved):

PR #425, opened 2021-04-18
Add sensors dynamically based on available device state

### LINKED ISSUE #223: [Architecture] Allow multiple entities per device

iMicknl:
Fixes #223

Screen Shot 2021-04-18 at 18 31 39


custom_components/tahoma/state_sensor.py
```diff
@@ -0,0 +1,105 @@
...
+    name: str
+    icon: Optional[str] = None
+    unit: Union[None, str, Callable[[dict], str]] = None
+    value: Callable[[Any], Any] = lambda val: val
+    device_class: Optional[str] = None
+    default_enabled: bool = True
+
+
+supported_states = {
+    "core:BatteryState": StateDescription(
```
tetienne, line 28: You don't use the key of this dict.

custom_components/tahoma/state_sensor.py
```diff
@@ -0,0 +1,105 @@
...
+        key="core:BatteryState",
```
tetienne, line 29: You can remove the key here and give it to the constructor. It will avoid to duplicate the data.

iMicknl: Would you know how to pass this?

custom_components/tahoma/__init__.py
```diff
@@ -151,6 +151,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
 
     platforms = defaultdict(list)
     platforms[SCENE] = scenarios
+    platforms["sensor"] = []
```
tetienne, line 154: No constant?

custom_components/tahoma/sensor.py
```diff
@@ -116,6 +117,22 @@ async def async_setup_entry(hass, entry, async_add_entities):
         if device.states
     ]
 
+    for platform, devices in data["platforms"].items():
+        if platform == "scene":
+            continue
+
+        for device in devices:
+            if device.states:
+                for state in device.states:
```
tetienne, line 126:
```suggestion
            for state in getattr(device, "states", []):
                
```

custom_components/tahoma/state_sensor.py
```diff
@@ -0,0 +1,151 @@
...
+    ),
+}
+
+
+class TahomaStateSensor(TahomaEntity, Entity):
+    """Representation of a TaHoma Sensor, based on a secondary device."""
+
+    def __init__(
+        self,
+        device_url: str,
```
tetienne, line 108: I would add the key as a parameter and remove it from the state_description to avoid the duplicate data. https://github.com/iMicknl/ha-tahoma/pull/425/files#r615452018

tetienne (commented):
Do you plan to clean the state_attributes method? https://github.com/iMicknl/ha-tahoma/blob/669ac3f1f5448e521cdd2223a6226bb06b319045/custom_components/tahoma/tahoma_entity.py#L80

tetienne:
I really like this PR :)

tetienne:
For my awning:

image

And my Energy meter:

image

For this last, that’s not really readable.

iMicknl:
> And my Energy meter (io:TotalElectricalEnergyConsumptionIOSystemSensor):
> 
> For this last, that’s not really readable.

Lol, I haven't seen a device like this yet. Maybe we should include the device name in the sensor name as well? Is your energy sensor one device, or is this part of our merge based on subsystemid (#1 + #2 etc.)

tetienne:
Yes that’s the most complex devices I own :D That’s indeed a device merging several Somfy device.
```

{JSON
      "creationTime": 1496693986000,
      "lastUpdateTime": 1496693986000,
      "label": "Atlantic",
      "deviceURL": "io://xxxxx#1",
      "shortcut": false,
      "controllableName": "io:TotalElectricalEnergyConsumptionIOSystemSensor",
      "definition": {
        "commands": [],
        "states": [
          {
            "type": "DiscreteState",
            "values": [
              "good",
              "low",
              "normal",
              "verylow"
            ],
            "qualifiedName": "core:DiscreteRSSILevelState"
          },
          {
            "type": "ContinuousState",
            "qualifiedName": "core:ElectricEnergyConsumptionState"
          },
          {
            "type": "ContinuousState",
            "qualifiedName": "core:RSSILevelState"
          },
          {
            "type": "DiscreteState",
            "values": [
              "dead",
              "lowBattery",
              "maintenanceRequired",
              "noDefect"
            ],
            "qualifiedName": "core:SensorDefectState"
          },
          {
            "type": "DiscreteState",
            "values": [
              "available",
              "unavailable"
            ],
            "qualifiedName": "core:StatusState"
          }
        ],
        "dataProperties": [],
        "widgetName": "CumulativeElectricPowerConsumptionSensor",
        "uiProfiles": [
          "ElectricEnergyConsumption"
        ],
        "uiClass": "ElectricitySensor",
        "qualifiedName": "io:TotalElectricalEnergyConsumptionIOSystemSensor",
        "type": "SENSOR"
      },
```

custom_components/tahoma/__init__.py
tetienne: You can just add the sensors platform manually once we retrieved all the available platforms.

iMicknl: > Do you plan to clean the state_attributes method?
> 
> https://github.com/iMicknl/ha-tahoma/blob/669ac3f1f5448e521cdd2223a6226bb06b319045/custom_components/tahoma/tahoma_entity.py#L80

I was thinking about removing this fully eventually, especially since this is not allowed in core. However, in the mean-time it can help us spot attributes / states that we should add.

iMicknl:
> Do you plan to clean the state_attributes method?
> 
> https://github.com/iMicknl/ha-tahoma/blob/669ac3f1f5448e521cdd2223a6226bb06b319045/custom_components/tahoma/tahoma_entity.py#L80

I was thinking about removing this fully eventually, especially since this is not allowed in core. However, in the mean-time it can help us spot attributes / states that we should add.

tetienne:
Do you think you will have some time to finish this PR. It's a nice improvement I would like to have.

iMicknl:
> Do you think you will have some time to finish this PR. It's a nice improvement I would like to have.

I will take some time in the coming weeks, however I could use some help as well :). Feel free to contribute / test as well.

tetienne:
Oki I will have a look too 😊

custom_components/tahoma/const.py
```diff
@@ -76,7 +76,6 @@
     "HitachiAirToAirHeatPump": CLIMATE,  # widgetName, uiClass is HitachiHeatingSystem (not supported)
     "HumiditySensor": SENSOR,
     "Light": LIGHT,
-    "LightSensor": SENSOR,
```
tetienne, line 79: TODO: don’t warn when LightSensor detected.

custom_components/tahoma/cover_devices/tahoma_cover.py
```diff
@@ -128,19 +125,6 @@ def is_closed(self):
 
         return None
 
-    @property
-    def icon(self):
```
tetienne, line 132: Replace By a sensor.

custom_components/tahoma/const.py
iMicknl, line 79: @tetienne should we already remove those sensors? Or keep this PR for secondary sensors and in a follow up rewrite our full sensor (and eventually binary sensor) system?

Having our sensor entities different will have other complications if we don't have a 'main' device/entity anymore.

custom_components/tahoma/cover_devices/tahoma_cover.py
```diff
@@ -257,17 +241,6 @@ def is_closing(self):
...
         )
 
-    @property
-    def device_state_attributes(self):
-        """Return the device state attributes."""
-        attr = super().device_state_attributes or {}
-
-        # Obstruction Detected attribute is used by HomeKit
-        if self.has_state(IO_PRIORITY_LOCK_LEVEL_STATE):
-            attr[ATTR_OBSTRUCTION_DETECTED] = True
```
iMicknl, line 267: Please don't remove this. Obstruction Detected is used by HomeKit and for now it needs to be a state attribute on the cover.

custom_components/tahoma/sensor.py
```diff
@@ -87,7 +84,7 @@
     "core:TemperatureInCelsius": TEMP_CELSIUS,
     "core:TemperatureInKelvin": TEMP_KELVIN,
     "core:TemperatureInFahrenheit": TEMP_FAHRENHEIT,
-    "core:LuminanceInLux": UNIT_LX,
+    "core:LuminanceInLux": LIGHT_LUX,
```
iMicknl, line 87: Don't forget to bump the minimum version of HA we need. Maybe we should just remove some other try catch import things soon and require the latest 2021.8 version.

custom_components/tahoma/state_sensor.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+SUPPORTED_STATES = [
+    StateDescription(
+        key="core:BatteryState",
+        name="Battery",
+        unit=PERCENTAGE,
+        device_class=sensor.DEVICE_CLASS_BATTERY,
+    ),
+    StateDescription(
```
iMicknl, line 41: I would love to use `SensorEntityDescription`, see https://github.com/home-assistant/core/pull/53890/files for an example.

We would still need to create our own `OverkizSensorDescription` since we need some additional things like the `value` transformer, but that is fine.

custom_components/tahoma/tahoma_entity.py
```diff
@@ -199,14 +199,6 @@ async def async_cancel_command(self, exec_id: str):
         """Cancel device command in async context."""
         await self.coordinator.client.cancel_command(exec_id)
 
-    def get_base_device_url(self):
```
iMicknl, line 202: Why is this removed?

custom_components/tahoma/tahoma_entity.py
```diff
@@ -81,9 +84,6 @@ def device_state_attributes(self) -> Dict[str, Any]:
         """Return the state attributes of the device."""
         attr = {}
 
-        if self.has_state(CORE_RSSI_LEVEL_STATE):
-            attr[ATTR_RSSI_LEVEL] = self.select_state(CORE_RSSI_LEVEL_STATE)
-
         if self.has_state(CORE_BATTERY_STATE):
```
iMicknl, line 87: This will be a harder one to migrate? Are there situations where we have two battery states? One with the number and one with just 'normal / low' etc? If that is never the case, we could migrate this one as well.

custom_components/tahoma/cover_devices/tahoma_cover.py
tetienne: That’s still the case ?

custom_components/tahoma/sensor.py
tetienne: Oh I didn’t know it was recently added. In which version it was released ?

custom_components/tahoma/cover_devices/tahoma_cover.py
iMicknl: https://github.com/home-assistant/core/blob/236738c455b827413ba052b1ea3bb24214d9f5db/homeassistant/components/homekit/type_covers.py#L177-L182

custom_components/tahoma/sensor.py
iMicknl: Nevermind, it has been added on Sep 23, 2020... I thought this was also part of a recent commit where they refactored some of these units.

https://github.com/home-assistant/core/commit/6c8e0e20fbd0da03cc5a5d2561a94a4721367bf5

custom_components/tahoma/state_sensor.py
tetienne: Oh clearly that’s something we should do with this PR. I totally missed this new class.

custom_components/tahoma/sensor.py
```diff
@@ -87,7 +84,7 @@
     "core:TemperatureInCelsius": TEMP_CELSIUS,
```
iMicknl, line 84: This will also make it harder by the way. If we move all sensor entities to this new structure, we cannot use this list easily. `"core:MeasuredValueType"` does only exist once for a device, where it can have multiple states. We can only use this when we know what the primary state is of a certain ui class/widget.

custom_components/tahoma/tahoma_entity.py
tetienne: I didn’t know what name gave to my function as I want to return the base url and the index. :S

iMicknl: `split_device_url`? Don't know either..

custom_components/tahoma/state_sensor.py
iMicknl: Done.

custom_components/tahoma/const.py
tetienne: So should I remove also my StateDescription related to the light?

custom_components/tahoma/tahoma_entity.py
tetienne: Ahah you see :D

custom_components/tahoma/tahoma_entity.py
```diff
@@ -3,42 +3,20 @@
...
-ATTR_RSSI_LEVEL = "rssi_level"
-
-CORE_AVAILABILITY_STATE = "core:AvailabilityState"
-CORE_BATTERY_STATE = "core:BatteryState"
 CORE_MANUFACTURER = "core:Manufacturer"
 CORE_MANUFACTURER_NAME_STATE = "core:ManufacturerNameState"
 CORE_MODEL_STATE = "core:ModelState"
 CORE_PRODUCT_MODEL_NAME_STATE = "core:ProductModelNameState"
-CORE_RSSI_LEVEL_STATE = "core:RSSILevelState"
-CORE_SENSOR_DEFECT_STATE = "core:SensorDefectState"
```
tetienne, line 23: We need this one. I just checked, and it is used when there is no more battery. For instance:

image

```JSON
{
	"states": [
		{
			"name": "core:StatusState",
			"type": 3,
			"value": "unavailable"
		},
		{
			"name": "core:SensorDefectState",
			"type": 3,
			"value": "dead"
		},
		{
			"name": "core:DiscreteRSSILevelState",
			"type": 3,
			"value": "verylow"
		},
		{
			"name": "core:RSSILevelState",
			"type": 2,
			"value": 4
		},
		{
			"name": "core:LuminanceState",
			"type": 2,
			"value": 110
		}
	]
}
```

Or perhaps do you plan to implement a battery sensor ?

custom_components/tahoma/cover_devices/tahoma_cover.py
tetienne: Done.

custom_components/tahoma/cover_devices/tahoma_cover.py
```diff
@@ -64,6 +64,8 @@
 ICON_LOCK_ALERT = "mdi:lock-alert"
```
iMicknl: Am I correct that we don't use these icons anymore?

custom_components/tahoma/tahoma_entity.py
iMicknl: Good to know! I didn't know that this was an important state. The issue is that battery state was coming from multiple properties...

What would you suggest? Add them as multiple sensors or create a separate battery sensor entity? My preference is to have multiple sensors, if the three states are not available at the same time in a device. Than 3 different sensors make sense. However I don't want to add 2 battery sensors accidentally for a device.

tetienne: I think a battery sensor will make more sense, and reduce code complexity. And it will avoid as you said to have multiple sensors linked to the same topic.

iMicknl: Agree! However, are we sure that the battery state can exist multiple times?

custom_components/tahoma/cover_devices/tahoma_cover.py
tetienne: @iMicknl About this one, I just checked, the state is returned only when the issue occured. Meaning we cannot create the sensor during the setup 99% of the time. What do you think about hardcoding it ?

iMicknl: @tetienne hmm, this is a hard one. Should we create an issue for this one and tackle it in a follow up PR? 
I think that `CORE_PRIORITY_LOCK_TIMER_STATE` is always present, where `IO_PRIORITY_LOCK_ORIGINATOR_STATE` is only present when there is a reason (lock timer state > 0).

custom_components/tahoma/tahoma_entity.py
iMicknl: @tetienne should we create an issue for this and do it in a follow up PR? Let's get this huge PR merged asap.

tetienne: Yep, I agree. Just roll back the battery code to be sure we still have the battery logic somewhere.

custom_components/tahoma/const.py
tetienne: At the end, we should ignore all the SENSOR devices. But yes, for the moment we can add it the ignore list.

custom_components/tahoma/tahoma_entity.py
iMicknl: Done.

custom_components/tahoma/const.py
```diff
@@ -39,13 +38,26 @@
...
+    "SunIntensitySensor",
+    "SunSensor",
+    "GasSensor",
+    "GenericSensor",
+    "HumiditySensor",
+    "TemperatureSensor",
+    "ThermalEnergySensor",
+    "WaterSensor",
+    "WeatherSensor",
+    "WindSensor",
```
tetienne, line 54: You can add LightSensor.

custom_components/tahoma/cover_devices/tahoma_cover.py
tetienne: Agreed. We can rollback this and focus on it later.

custom_components/tahoma/sensor.py
```diff
@@ -1,116 +1,338 @@
...
+    OverkizSensorDescription(
+        key="core:BatteryState",
+        name="Battery",
+        device_class=sensor.DEVICE_CLASS_BATTERY,
+        value=lambda value: str(value).capitalize(),
+    ),
+    OverkizSensorDescription(
+        key="core:RSSILevelState",
+        name="RSSI Level",
+        value=lambda value: round(value),
```
tetienne, line 61: Why round the value for every sensor? The raw data is not correct?

custom_components/tahoma/sensor.py
```diff
@@ -1,116 +1,338 @@
...
+        unit_of_measurement=SIGNAL_STRENGTH_DECIBELS,
+        device_class=sensor.DEVICE_CLASS_SIGNAL_STRENGTH,
+        state_class=STATE_CLASS_MEASUREMENT,
+    ),
+    OverkizSensorDescription(
+        key="core:ExpectedNumberOfShowerState",
```
tetienne, line 67: This one is magic!

custom_components/tahoma/sensor.py
```diff
@@ -1,116 +1,338 @@
...
+        key="io:InletEngineState",
+        name="Inlet Engine",
+        icon="mdi:fan-chevron-up",
+        value=lambda value: round(value),
+        unit_of_measurement=VOLUME_FLOW_RATE_CUBIC_METERS_PER_HOUR,
+        state_class=STATE_CLASS_MEASUREMENT,
+    ),
+    OverkizSensorDescription(
+        key="hlrrwifi:RoomTemperatureState",
+        name="Room Temperature",
```
tetienne, line 116: To ease the code, we can generate the name from the key.

custom_components/tahoma/sensor.py
```diff
@@ -1,116 +1,338 @@
...
+        icon="mdi:alert",
+    ),
+    OverkizSensorDescription(
+        key="core:FossilEnergyConsumptionState",
+        name="Fossil Energy Consumption",
+        device_class=sensor.DEVICE_CLASS_ENERGY,
+    ),
+    OverkizSensorDescription(
+        key="core:GasConsumptionState",
+        name="Gas Consumption",
```
tetienne, line 140: No device class?

custom_components/tahoma/sensor.py
```diff
@@ -122,68 +344,59 @@ async def async_setup_entry(
...
-            CORE_LUMINANCE_STATE,
-            CORE_RELATIVE_HUMIDITY_STATE,
-            CORE_SUN_ENERGY_STATE,
-            CORE_TEMPERATURE_STATE,
-            CORE_THERMAL_ENERGY_CONSUMPTION_STATE,
-            CORE_WINDSPEED_STATE,
-            CORE_WATER_CONSUMPTION_STATE,
-        )
-        return round(state, 2) if state is not None else None
+        state = self.device.states[self.entity_description.key]
```
tetienne, line 384: We can do the same for the binary sensor pr.

custom_components/tahoma/sensor.py
iMicknl: afaik, there is no device class for gas consumption sensors.

custom_components/tahoma/sensor.py
iMicknl: I thought about this as well, however I don't see the value. This lookup is fast and gives us a way to easily change the name.

We can always add this in a follow up PR? However, I think there are more pressing issues to solve :-).

custom_components/tahoma/sensor.py
iMicknl: Tbh, I don't recall why I added this. There are quite some states that give very precise measures, thus I chose to round it. 

Should I remove all the rounds and add them back based on user reports? Or should we keep it safe and just keep them?

custom_components/tahoma/const.py
iMicknl: Done

custom_components/tahoma/cover_devices/tahoma_cover.py
iMicknl: Done

custom_components/tahoma/sensor.py
tetienne: Sure. It’s just I like to reduce number of lines :D

custom_components/tahoma/sensor.py
tetienne: I would suggest to keep the raw value. And see if user complains.

tetienne (dismissed):
Tested locally. Seems to be OK. I can even now add my energy sensors within the Energy panel.

custom_components/tahoma/sensor.py
iMicknl: Done! I just added it for RSSI, since I can test this one and it always return .0, thus rounding will make it cleaner.

tetienne (approved):

PR #426, opened 2021-04-18
Reduce battery state complexity

tetienne:


iMicknl (approved):

vlebourl (approved):

iMicknl:
@tetienne this part is by the way not used in core, since we are not allowed to have this captured in attributes. It needs to become a separate battery sensor..

tetienne:
Damn, my first PR since a long time, and it's useless :')

iMicknl:
@tetienne nah, we can use this in https://github.com/iMicknl/ha-tahoma/pull/425! So this is still a good improvement <3.

PR #444, opened 2021-05-18
Fix broken tests

tetienne:
Because we override an existing component, the tests are still failing. The legacy component is loaded instead of ours.


tetienne:
@iMicknl Tests are ok now 👌

vlebourl (approved):
I trust you ^^'

PR #448, opened 2021-05-21
use enable_custom_integrations

MatthewFlamm:
Sorry for the second PR, I targeted the wrong branch of the first one.  Meant to target your branch from #444 .

This PR home-assistant/core#43692 broke your usage of loading custom integrations, but I'm not quite sure why it didn't break _all_ usage of custom integrations in tests.


tetienne:
@MatthewFlamm That's awesome! 🥳 I didn't know about this fixture. Good catch about the Core PR. Perhaps the other components are not impacted because they don't override an existing component 🤷‍♂️

PR #408, opened 2021-03-17
fixes for "Add support for AtlanticElectricalHeaterWithAdjustableTemperatureSetpoint"

tillstaff:
From PR #362 : 
- fix broken io:AtlanticElectricalHeaterIOComponent device
- fix preset modes
- need refactoring to handle prog mode and auto mode as preset

PR #454, opened 2021-05-28
Adjust Atlantic Electrical Heater to new standards and fix issues

iMicknl:
As mentioned by @tillstaff in, there are some bugs in the Atlantic Electrical Heater implementation.

https://github.com/iMicknl/ha-tahoma/pull/408


custom_components/tahoma/climate_devices/atlantic_electrical_heater.py
```diff
@@ -52,35 +52,31 @@ def supported_features(self) -> int:
...
     @property
     def hvac_modes(self) -> List[str]:
         """Return the list of available hvac operation modes."""
-        return [*MAP_HVAC_MODES]
+        return [*HVAC_MODES_TO_TAHOMA]
+
+    async def async_set_hvac_mode(self, hvac_mode: str) -> None:
+        """Set new target hvac mode."""
+        await self.async_execute_command(
+            COMMAND_SET_HEATING_LEVEL, HVAC_MODES_TO_TAHOMA[hvac_mode]
```
tillstaff, line 65: This command won't work for the PASS_ACTUATOR device.

Here are the commands sent by the Cozytouch application : 

Set to Off (the command is named "off" without parameters): 
```
2021-06-01 10:22:50 INFO (MainThread) [custom_components.tahoma] 2021-06-01 10:21:12: off executed via Android 2.5.4#244 Set heating level in room c423fb05-92c5-49b4-a917-c7ac553e7ef4 on io://xxxx-xxxx-xxxx/2634956, with [].
```

Set to On (there is no command for "on", it's just set the preset to "comfort"): 
```
2021-06-01 10:22:50 INFO (MainThread) [custom_components.tahoma] 2021-06-01 10:21:37: setHeatingLevel executed via Android 2.5.4#244 Set heating level in room c423fb05-92c5-49b4-a917-c7ac553e7ef4 on io://xxxx-xxxx-xxxx/2634956, with ['comfort'].
```

custom_components/tahoma/climate_devices/atlantic_electrical_heater.py
```diff
@@ -18,22 +18,22 @@
 
 COMMAND_SET_HEATING_LEVEL = "setHeatingLevel"
 
+CORE_ON_OFF_STATE = "core:OnOffState"
 IO_TARGET_HEATING_LEVEL_STATE = "io:TargetHeatingLevelState"
 
-PRESET_FREEZE = "Freeze"
-PRESET_FROST_PROTECTION = "frostprotection"
-PRESET_OFF = "off"
+PRESET_FREEZE = "freeze"
```
tillstaff, line 24: Why did you rename "frost protection" to "freeze"? Frostprotection seems well named but I could be wrong.
In french, the function is named "hors-gel" and it will just set the temperature of the heater to 7°C to avoid any freezing in the house.

Although, the following presets are needed too in the Atlantic_Electrical_Heater device : 
PRESET_COMFORT1 = "comfort-1"
PRESET_COMFORT2 = "comfort-2"

iMicknl: I didn't change it, in this version it wasn't fixed yet. I will fix it was part of this PR as well!

custom_components/tahoma/climate_devices/atlantic_electrical_heater.py
iMicknl: Off should work already in this PR, but I understood that I have to rename `on` to `comfort`?

custom_components/tahoma/climate_devices/atlantic_electrical_heater.py
tillstaff: Ok I see, I thought the commit was coming from the other PR, sorry.

So just the two presets are missing.

custom_components/tahoma/climate_devices/atlantic_electrical_heater.py
tillstaff: I just test it and yes, it's working for "off" (it's not the same command sent by the cozytouch app but it doesn't matter).

It will work ok with "comfort", but you need to keep the "on" value somewhere for the "hvac_mode(self)" function.

PR #458, opened 2021-06-04
Add TaHoma switch option

ahivert:
After reverse engineering the android app, I found out that the endpoint used by "TaHoma switch" was the same as eedomus one.

This PR add "TaHoma switch" option to the UI and so let user select it. User have to enter same login information that he use on the app.

Tested on my side with protocols io and rts (thank you!).

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ Custom component for Home Assistant to interact with smart devices via the Somfy
 ## Supported hubs
 
 - Somfy TaHoma Box
+- Somfy TaHoma switch
 - Somfy Connexoon IO
 - Somfy Connexoon RTS
 - Cozytouch
--- a/custom_components/tahoma/const.py
+++ b/custom_components/tahoma/const.py
@@ -21,6 +21,7 @@ SUPPORTED_ENDPOINTS = {
     "Somfy Connexoon IO": "https://tahomalink.com/enduser-mobile-web/enduserAPI/",
     "Somfy Connexoon RTS": "https://ha201-1.overkiz.com/enduser-mobile-web/enduserAPI/",
     "Somfy TaHoma": "https://tahomalink.com/enduser-mobile-web/enduserAPI/",
+    "Somfy TaHoma switch": "https://ha101-1.overkiz.com/enduser-mobile-web/enduserAPI/",
 }
 
 HUB_MANUFACTURER = {
@@ -31,6 +32,7 @@ HUB_MANUFACTURER = {
     "Somfy Connexoon IO": "Somfy",
     "Somfy Connexoon RTS": "Somfy",
     "Somfy TaHoma": "Somfy",
+    "Somfy TaHoma switch": "Somfy",
 }
 
 MIN_UPDATE_INTERVAL = 30


iMicknl:
Thanks for your first contribution @ahivert, much appreciated! Do you own a Somfy TaHoma Switch, or did you just base it on the app?

The eedomus entry is faulty unfortunately and will be removed. In our PR to core we use an improved list https://github.com/home-assistant/core/blob/c528645335dfa204ce837d1629548c8070f2d927/homeassistant/components/overkiz/const.py#L15-L23. Apparently Somfy uses different servers per region, not per device. I believe that `https://ha101-1.overkiz.com` would be the same as `tahomalink.com`, but I haven't been able to test that yet.

ahivert:
Yes I own the TaHoma switch and can do more test if needed.

iMicknl:
> Yes I own the TaHoma switch and can do more test if needed.

I see you are in the Europe region, have you tried the normal Somfy TaHoma entry? (which does use tahomalink.com). And do you have a webinterface as well, or can you only use the app?

Cool to see a user of the switch, I didn't know it was already available for purchase!

ahivert:
Unfortunately we can't use `tahomalink.com` with the "TaHoma switch". We can only control it through the new ios/android apps names "TaHoma" :cry: .

image

tetienne:
So there is no web interface? What about https://www.somfy-connect.com/fr/?

ahivert:
Yes, no web interface :cry: 

image

I find out after I received it, HA become a must have with this box.

If you can read french, a guy from somfy support explain that `tahomalink.com` is not available with this box (fourth answer):
https://forum.somfy.fr/questions/2696600-impossible-faire-fonctionner-tahoma-switch

tetienne:
@ahivert Sorry again to ask a question not related to this topic, but can you try the official Somfy component within HA I’m maintain too. I want to know if their official API is well supported by the Tahoma Switch.

tetienne (approved):

ahivert:
> @ahivert Sorry again to ask a question not related to this topic, but can you try the official Somfy component within HA I’m maintain too. I want to know if their official API is well supported by the Tahoma Switch.

Already tried, it looks like it authenticate me but then nothing is added in devices list (on HA).

iMicknl:
@ahivert apologies for the late reply. Did you also try configuring this integration with the 'Somfy TaHoma' choice? HomeBridge users reported that this works as well, so I wonder why this didn't work in your case.

ahivert:
> @ahivert apologies for the late reply. Did you also try configuring this integration with the 'Somfy TaHoma' choice? HomeBridge users reported that this works as well, so I wonder why this didn't work in your case.

Yes tried that first but got an error. I then tried to investigate and find out that the endpoint of the API used by the android app was the same one as eedomus so I selected eedomus.

But now that's strange I've reset home assistant integration to select somfy tahoma and now it works 😅

Looks like we can close this issue/PR.

Thanks ! 🙂

iMicknl:
Thanks for double checking @ahivert! The endpoint you found (thanks by the way!) is the same as tahomalink.com redirects to in the back-end, thus my confusion. Great to hear that it works now. :)

And thanks for your contribution. If you want to contribute more, let us know ;-). https://github.com/iMicknl/ha-tahoma/issues/471

Issue #459, opened 2021-06-13
Update state and attributes when event received from TahomaClient


laszlojakab:
- [x] **I have read the [README](https://github.com/iMicknl/ha-tahoma/blob/master/README.md), including the [Advanced](https://github.com/iMicknl/ha-tahoma/blob/master/README.md#advanced) section regarding debugging.**

**Is your feature request related to a problem? Please describe.**
I'm developing a lovelace card for my Somfy raffstores, and would like to sync the animation of it as the raffstore goes up or down. 
The 30 seconds update of the integration is terrible rear for it. In the worst case when I open one of my raffstore it takes 30 seconds to update the image in the raffstore card (in that time in reality the raffstore is 2/3 part opened at that time). I checked in pyhoma the TahomaClient.fetch_events method what happens when I open a raffstore. When I opened the raffstore an event almost immediatly fired with the following device states:
* core:MovingState=true
* core:TargetClosureState=0

When the raffstore fully opened another event also received immediatly with the following state:
* core:MovingState, false

When the first time the raffstore opens I could measure the time (from code) between the two events and next time I can animate raffstore image synchronized with reality.

Can integration update cover state and attributes when an event arrives from TahomeClient?

**Describe the solution you'd like**
Update HA cover state and attributes when event received from TahomaClient.

**Additional context**
Maybe `should_poll` has to be set to false in cover entity (possibly in all tahomaentity) and `async_schedule_update_ha_state` function should be used to force HA to update state when event received from TahomaClient. 
As I see integration uses DataUpdateCoordinator. I have never used DataUpdateCoordinator in my integrations so maybe `async_schedule_update_ha_state` won't work in this architecture.

tetienne:
Actually what you described is already what we do. See

tetienne:
What you described is already what we do. See https://github.com/iMicknl/ha-tahoma/blob/master/custom_components/tahoma/coordinator.py

Can you check the event you see is well of type device changed?

If yes, it probably means we have to modify cover.py to handle the state variable you gave.

laszlojakab:
OK there are multiple issues:
- I used position slider to set position to 0 (close) or to 100 (open). In that way state never changes to closing or opening. Can integration change this behaviour? If target position is 0 then set the state to closing, and if target position is 100 than set state to opening during the operation.
- The recorder logs of closing the raffstore (by pressing close button, not setting position to 0), contains an "open" state which is missleading (it should be closing):

| timestamp | state | attributes |
[...]

PR #460, opened 2021-06-13
Support is_opening/closing when moving position

tetienne:
Fix https://github.com/iMicknl/ha-tahoma/issues/459

### DIFF
--- a/custom_components/tahoma/cover.py
+++ b/custom_components/tahoma/cover.py
@@ -58,6 +58,7 @@ CORE_CLOSURE_STATE = "core:ClosureState"
 CORE_CLOSURE_OR_ROCKER_POSITION_STATE = "core:ClosureOrRockerPositionState"
 CORE_DEPLOYMENT_STATE = "core:DeploymentState"
 CORE_MEMORIZED_1_POSITION_STATE = "core:Memorized1PositionState"
+CORE_MOVING_STATE = "core:MovingState"
 CORE_OPEN_CLOSED_PARTIAL_STATE = "core:OpenClosedPartialState"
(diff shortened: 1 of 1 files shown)


custom_components/tahoma/cover.py
```diff
@@ -317,20 +317,50 @@ async def async_my(self, **_):
...
+        if self.is_awning():
+            # TODO
+            return False
+
+        if self.select_state(CORE_CLOSURE_STATE) > self.select_state(
+            CORE_TARGET_CLOSURE_STATE
+        ):
+            return True
+
+        return False
```
tetienne, line 339: Quick and dirty, need to be clean.

tetienne:
@iMicknl It’s ok for you?

vlebourl (approved):

custom_components/tahoma/cover.py
```diff
@@ -317,19 +317,41 @@ async def async_my(self, **_):
...
         """Return if the cover is opening or not."""
-        return any(
+        if any(
             execution.get("deviceurl") == self.device.deviceurl
             and execution.get("command_name") in COMMANDS_OPEN + COMMANDS_OPEN_TILT
             for execution in self.coordinator.executions.values()
+        ):
+            return True
+
+        is_moving = self.device.states[CORE_MOVING_STATE].value
```
iMicknl, line 327: Do we still need the loop through current executions, when we can look at the state as well?

tetienne: I thought about it too. But I don’t know how it will work for the tilt. It’s why I keep it like this.

iMicknl (approved):

Issue #479, opened 2021-07-15
Improve typing of this integration


iMicknl:
### Did you read the instructions?

- [X] I have read the [README](https://github.com/iMicknl/ha-tahoma/blob/master/README.md), including the [advanced debugging](https://github.com/iMicknl/ha-tahoma/blob/master/README.md#advanced) section.

### The request

Since HA core has moved to strict typechecks, they have introduced many nice types. If we want to add this integration to core eventually, we will need to have the integration typed as well.

tetienne:
We should also add mypy as pre commit.

iMicknl:
> We should also add mypy as pre commit.

Maybe we can inherit the typings from core already, but that will require some more PR’s haha.

iMicknl:
```
homeassistant/components/overkiz/coordinator.py:74:5: C901 'OverkizDataUpdateCoordinator._async_update_data' is too complex (26)
```

```
homeassistant/components/overkiz/cover_devices/vertical_cover.py:56: error: Class cannot subclass 'OverkizGenericCover' (has type 'Any')  [misc]
homeassistant/components/overkiz/cover_devices/vertical_cover.py:60: error: Function is missing a return type annotation  [no-untyped-def]
homeassistant/components/overkiz/cover_devices/vertical_cover.py:79: error: Function is missing a return type annotation  [no-untyped-def]
homeassistant/components/overkiz/cover_devices/vertical_cover.py:88: error: Function is missing a return type annotation  [no-untyped-def]
homeassistant/components/overkiz/cover_devices/vertical_cover.py:106: error: Function is missing a type annotation  [no-untyped-def]
homeassistant/components/overkiz/cover_devices/vertical_cover.py:108: error: Call to untyped function "is_low_speed_enabled" in typed context  [no-untyped-call]
homeassistant/components/overkiz/cover_devices/vertical_cover.py:114: error: Function is missing a type annotation  [no-untyped-def]
homeassistant/components/overkiz/cover_devices/vertical_cover.py:116: error: Call to untyped function "is_low_speed_enabled" in typed context  [no-untyped-call]
homeassistant/components/overkiz/cover_devices/vertical_cover.py:123: error: Function is missing a type annotation  [no-untyped-def]
homeassistant/components/overkiz/cover_devices/vertical_cover.py:125: error: Call to untyped function "is_low_speed_enabled" in typed context  [no-untyped-call]
[...]

PR #480, opened 2021-07-15
Improve typing for `async_setup_entry`

iMicknl:
Part of #479.


tetienne (approved):

tetienne:
You run mypy to check ?

iMicknl:
> You run mypy to check ?

These I checked, but would be good to have mypy strict checks added to the whole integration in another PR.

PR #494, opened 2021-07-23
Added German translation

azrael783:
I hope that the name of the file fits. Is there something else which needs to be done or will the translation now work out of the box?


tetienne (approved):

tetienne:
Thx for this PR. Normally it should be enough. You didn't try locally?

azrael783:
I tried, but I have never worked with remote containers before. So I have no glue what I have to do ...

iMicknl:
Thanks for your first PR, @azrael783! I will have a look, probably after this weekend, but it looks good.

Normally, you wouldn't have to translate all strings, but some of them will be 'translated' by Home Assistant Lokalise. We use certain keys in strings.json (see https://github.com/iMicknl/ha-tahoma/blob/master/custom_components/tahoma/strings.json) and they will be automatically translated to match the other Home Assistant integrations.

This is no issue for this custom integration, but eventually it would be nice to generate them with Hassfest or something and just translate the custom strings. But let's discuss on Discord sometime :-).

iMicknl:
Thanks!

PR #496, opened 2021-07-26
Add My Home Assistant button to README

iMicknl:


README.md
```diff
@@ -22,29 +22,31 @@ Custom component for Home Assistant to interact with smart devices via the Somfy
...
 
 ### Manual
 
-Copy the `custom_components/tahoma` to your `custom_components` folder. Reboot Home Assistant and install the Somfy TaHoma integration via the integrations page.
+Copy the `custom_components/tahoma` to your `custom_components` folder. Reboot Home Assistant and install the Somfy TaHoma integration via the integrations page or press the blue button below.
 
 ### HACS
 
-This integration is included in HACS. Search for the `Somfy TaHoma` integration and choose install. Reboot Home Assistant and install the Tahoma integration via the integrations page.
+This integration is included in HACS. Search for the `Somfy TaHoma` integration and choose install. Reboot Home Assistant and install the Somfy TaHoma integration via the integrations page or press the blue button below.
```
tetienne, line 34: Would be nice to move this at the top, to make the button more visible.

iMicknl: What do you mean? The blue button only works after you did install the integration manually or via HACS and after a reboot. Thus I believe we should not list it earlier than our install instructions.

vlebourl: What happens when you click the button without the component installed? I tend to agree with @tetienne that it would be nice to have it on top, even if it doesn't work when not installed.

tetienne: My bad I misread the button. I thought it was to install our component.

README.md
```diff
@@ -23,29 +23,31 @@ Custom component for Home Assistant to interact with smart devices via the Somfy
...
 
 ## Advanced
 
+### TaHoma services
+
+After installation this integration adds new services to Home Assistant which can be used in automations. The new services are:
+
++ Somfy TaHoma: Set cover position with low speed (tahoma.set_cover_position_low_speed, only added if supported by the device)
++ Somfy TaHoma: My position (cover) (tahoma.set_cover_my_position)
++ Somfy TaHoma: Execute command (tahoma.execute_command)
```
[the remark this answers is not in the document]
iMicknl: We could add this (in another PR), but I am also not a huge fan of the `execute_command` feature. Imo all these features should just be mapped to HA features or services, and not a global execute command feature.

README.md
iMicknl: > What happens when you click the button without the component installed? I tend to agree with @tetienne that it would be nice to have it on top, even if it doesn't work when not installed.

It will just give an error (which is not even very nice).

tetienne (approved):

tetienne:
You will rethink your PR, or it’s ok?

iMicknl:
> You will rethink your PR, or it’s ok?

Let's merge as is, and happy to merge follow up PR's with better suggestions :-)

PR #497, opened 2021-07-26
Add low speed switch entity

tetienne:
image

image

*Discret* is the french word for low mode

Instead of a service, I suggest to use an entity switch as device option. So no need anymore to create cover template.

For the icon, I choose the same one used by the Somfy switch: https://www.somfy.fr/produits/1033114/s-so-rs100-io

### DIFF
--- a/custom_components/tahoma/__init__.py
+++ b/custom_components/tahoma/__init__.py
@@ -209,7 +209,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     supported_platforms = set(platforms.keys())
 
     # Sensor and Binary Sensor will be added dynamically, based on the device states
-    # Required to load the low speed switch
+    # Switch will be added dynamically, based on device features (e.g. low speed cover switch)
     supported_platforms.update((BINARY_SENSOR, SENSOR, SWITCH))
     for platform in supported_platforms:
         hass.async_create_task(
--- a/custom_components/tahoma/__init__.py
+++ b/custom_components/tahoma/__init__.py
@@ -9,6 +9,7 @@ from aiohttp import ClientError, ServerDisconnectedError
 from homeassistant.components.binary_sensor import DOMAIN as BINARY_SENSOR
 from homeassistant.components.scene import DOMAIN as SCENE
 from homeassistant.components.sensor import DOMAIN as SENSOR
+from homeassistant.components.switch import DOMAIN as SWITCH
 from homeassistant.config_entries import SOURCE_IMPORT, ConfigEntry
 from homeassistant.const import CONF_EXCLUDE, CONF_PASSWORD, CONF_SOURCE, CONF_USERNAME
 from homeassistant.core import HomeAssistant, ServiceCall
@@ -208,8 +209,10 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     supported_platforms = set(platforms.keys())
 
     # Sensor and Binary Sensor will be added dynamically, based on the device states
+    # Required to load the low speed switch
     supported_platforms.add(BINARY_SENSOR)
     supported_platforms.add(SENSOR)
+    supported_platforms.add(SWITCH)
 
     for platform in supported_platforms:
         hass.async_create_task(
--- a/custom_components/tahoma/__init__.py
+++ b/custom_components/tahoma/__init__.py
@@ -210,10 +210,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
 
     # Sensor and Binary Sensor will be added dynamically, based on the device states
     # Required to load the low speed switch
-    supported_platforms.add(BINARY_SENSOR)
-    supported_platforms.add(SENSOR)
-    supported_platforms.add(SWITCH)
-
+    supported_platforms.update((BINARY_SENSOR, SENSOR, SWITCH))
     for platform in supported_platforms:
         hass.async_create_task(
             hass.config_entries.async_forward_entry_setup(entry, platform)
--- a/custom_components/tahoma/cover_devices/tahoma_cover.py
+++ b/custom_components/tahoma/cover_devices/tahoma_cover.py
@@ -1,5 +1,6 @@
 """Base class for TaHoma covers, shutters, awnings, etc."""
 import logging
+
 from homeassistant.components.cover import (
     ATTR_POSITION,
     ATTR_TILT_POSITION,
--- a/custom_components/tahoma/cover_devices/tahoma_cover.py
+++ b/custom_components/tahoma/cover_devices/tahoma_cover.py
@@ -1,4 +1,5 @@
 """Base class for TaHoma covers, shutters, awnings, etc."""
+import logging
 from homeassistant.components.cover import (
     ATTR_POSITION,
     ATTR_TILT_POSITION,
@@ -75,6 +76,8 @@ SERVICE_COVER_POSITION_LOW_SPEED = "set_cover_position_low_speed"
 SUPPORT_MY = 512
 SUPPORT_COVER_POSITION_LOW_SPEED = 1024
 
+_LOGGER = logging.getLogger(__name__)
+
 
 class TahomaGenericCover(OverkizEntity, CoverEntity):
     """Representation of a TaHoma Cover."""
@@ -92,6 +95,10 @@ class TahomaGenericCover(OverkizEntity, CoverEntity):
 
     async def async_set_cover_position_low_speed(self, **kwargs):
         """Move the cover to a specific position with a low speed."""
+        _LOGGER.warning(
+            "This service will removed in the next releases. Use instead the low speed switch."
+        )
+
         position = 100 - kwargs.get(ATTR_POSITION, 0)
 
         await self.executor.async_execute_command(
--- a/custom_components/tahoma/cover_devices/tahoma_cover.py
+++ b/custom_components/tahoma/cover_devices/tahoma_cover.py
@@ -97,7 +97,7 @@ class TahomaGenericCover(OverkizEntity, CoverEntity):
     async def async_set_cover_position_low_speed(self, **kwargs):
         """Move the cover to a specific position with a low speed."""
         _LOGGER.warning(
-            "This service will removed in the next releases. Use instead the low speed switch."
+            "This service will be removed in Setptember 2021. Use instead the low speed switch."
         )
 
         position = 100 - kwargs.get(ATTR_POSITION, 0)
--- a/custom_components/tahoma/cover_devices/tahoma_cover.py
+++ b/custom_components/tahoma/cover_devices/tahoma_cover.py
@@ -1,6 +1,4 @@
 """Base class for TaHoma covers, shutters, awnings, etc."""
-import logging
-
 from homeassistant.components.cover import (
     ATTR_POSITION,
     ATTR_TILT_POSITION,
@@ -77,8 +75,6 @@ SERVICE_COVER_POSITION_LOW_SPEED = "set_cover_position_low_speed"
 SUPPORT_MY = 512
 SUPPORT_COVER_POSITION_LOW_SPEED = 1024
 
-_LOGGER = logging.getLogger(__name__)
-
 
 class TahomaGenericCover(OverkizEntity, CoverEntity):
     """Representation of a TaHoma Cover."""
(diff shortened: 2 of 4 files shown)


custom_components/tahoma/tahoma_entity.py
```diff
@@ -128,7 +128,7 @@ def device_info(self) -> Dict[str, Any]:
 
         return {
             "identifiers": {(DOMAIN, self.base_device_url)},
-            "name": self.name,
+            "name": self.device.label,
```
tetienne, line 131: If I don’t do this, HA will take the name of the first loaded entity of the device. In my case, the switch.

custom_components/tahoma/cover_devices/tahoma_cover.py
```diff
@@ -92,6 +96,10 @@ def current_cover_tilt_position(self):
 
     async def async_set_cover_position_low_speed(self, **kwargs):
         """Move the cover to a specific position with a low speed."""
+        _LOGGER.warning(
+            "This service will removed in the next releases. Use instead the low speed switch."
```
iMicknl, line 100: Should we just mention a release, or even better a month? This service will be removed in September 2021 e.g.

tetienne: Thx for the suggestion. I didn’t know how to formulate. 
>This service will be removed in September 2021.
sounds good. @vlebourl what do you think?

custom_components/tahoma/cover_devices/vertical_cover.py
```diff
@@ -98,18 +99,38 @@ def current_cover_position(self):
...
+        if self.is_low_speed_enabled():
+            await self.async_execute_command(
+                COMMAND_SET_CLOSURE_AND_LINEAR_SPEED, 100, "lowspeed"
+            )
+        else:
+            await self.async_execute_command(self.select_command(*COMMANDS_CLOSE))
+
+    def is_low_speed_enabled(self):
+        """Return if low speed mode is enabled."""
+        switch_entity_id = f"{self.entity_id.replace('cover', 'switch')}_low_speed"
```
iMicknl, line 134: I am not a huge fan of this implementation, however this is most likely the easiest / fastest implementation. However I have seen that entity id's are not unique and if another integration tries to register the same one, it could be that you have something like `cover_low_speed_2` or something?

We could pull the entity id from registry based on the tahoma device id, on startup, however this is a more 'expensive' call.

custom_components/tahoma/switch.py
```diff
@@ -112,3 +127,46 @@ async def async_toggle(self, **_):
     def is_on(self):
         """Get whether the switch is in on state."""
         return self.select_state(CORE_ON_OFF_STATE, IO_FORCE_HEATING_STATE) == STATE_ON
+
+
+class TahomaLowSpeedSwitch(TahomaEntity, SwitchEntity, RestoreEntity):
```
iMicknl, line 132:
```suggestion
class TahomaLowSpeedCoverSwitch(TahomaEntity, SwitchEntity, RestoreEntity):
```

Maybe add cover to the name?

custom_components/tahoma/switch.py
```diff
@@ -112,3 +127,46 @@ async def async_toggle(self, **_):
...
+        """Initialize the low speed switch."""
+        super().__init__(device_url, coordinator)
+        self._is_on = False
+
+    async def async_added_to_hass(self):
+        """Run when entity about to be added."""
+        await super().async_added_to_hass()
+        state = await self.async_get_last_state()
+        if state:
+            self._is_on = state.state == STATE_ON
```
iMicknl, line 147: Looks like a good fix, but is there no helper to restore the state? I would say that this is something that RestoreEntity should have already. Or do they only offer `async_get_last_state`?

iMicknl (commented):
Nice PR, @tetienne! This is a lot nicer than the template covers. The only downside I see is that users cannot choose to mix the functionality anymore, without changing the switch.

E.g. if they want to open covers fast in one automation and open them silently in another one. However, I have some feeling that this is an edge case that no one wants really. :-)

custom_components/tahoma/cover_devices/vertical_cover.py
tetienne: I got your point. Here I retrieve the current entity id of the cover I want to manipulate. So for sure, it will be unique. 
But you think, it can still be an issue?
If you have the Somfy API integration, and this one, you will have two covers: `cover.my_cover` and `cover.my_cover_2`. If the latest is from this integration, I will create a `switch.my_cover_2_low_speed`.

tetienne: In parallel, I will submit the same feature to the core within the official integration. I will see what they suggest :D

custom_components/tahoma/switch.py
tetienne: https://github.com/home-assistant/core/blob/dev/homeassistant/helpers/restore_state.py#L241
That was my first though too. But after a discussion on Discord, and some research within the Core, it seems the correct pattern.

tetienne: For instance: https://github.com/home-assistant/core/blob/a59460a23336627d0bc12b1eefffdaa516e55e87/homeassistant/components/switchbot/switch.py#L46

tetienne:
> Nice PR, @tetienne! This is a lot nicer than the template covers. The only downside I see is that users cannot choose to mix the functionality anymore, without changing the switch.
>
> E.g. if they want to open covers fast in one automation and open them silently in another one. However, I have some feeling that this is an edge case that no one wants really. :-)

You are right. That’s another step in the automation. For this edge case, HA should support natively cover speed, or end user should create a template cover.

custom_components/tahoma/cover_devices/vertical_cover.py
iMicknl: Makes sense, let's keep it for now and we can always revisit in the future!

custom_components/tahoma/switch.py
iMicknl: Weird, but looks ok :-)

iMicknl:
> > Nice PR, @tetienne! This is a lot nicer than the template covers. The only downside I see is that users cannot choose to mix the functionality anymore, without changing the switch.
> > E.g. if they want to open covers fast in one automation and open them silently in another one. However, I have some feeling that this is an edge case that no one wants really. :-)
> 
> You are right. That’s another step in the automation. For this edge case, HA should support natively cover speed, or end user should create a template cover.

They can't create a template cover anymore if we remove the service :-). That was my point haha, however I am fine by removing the service (or at least add the warning) and see if people complain.

azrael783:
> 
> 
> Nice PR, @tetienne! This is a lot nicer than the template covers. The only downside I see is that users cannot choose to mix the functionality anymore, without changing the switch.
> 
> E.g. if they want to open covers fast in one automation and open them silently in another one. However, I have some feeling that this is an edge case that no one wants really. :-)

I think it still should be possible. The only thing you have to do is: activate or deactivate the switch prior to the moving of the cover.
But I was about to create buttons to move to a certain position with low speed and I am not sure if the `custom:button-card` supports two different service calls in once (activating the switch and then moving the cover).

tetienne:
@azrael783 For lovelace, on your card you can add a switch too. It will mimic Somfy interface.
Sometimes, you want to trigger the slow mode and sometimes not?

azrael783:
> 
> 
> @azrael783 For lovelace, on your card you can add a switch too. It will mimic Somfy interface.
> Sometimes, you want to trigger the slow mode and sometimes not?

I didn't get the first statement.
Yes, for example: in the morning and in the evening I want to open and close silently via an automation. But I want to have also fully control via the UI. As I understood the pull request will add for every cover a switch which allows me to activate the "silent" mode. So bevor I can move the cover with low speed I have to activate the switch, right? For me it makes then more sense to create templates for the UI because I think it is easier to just call the template rather then first activate the switch and then move the cover ...

tetienne:
@azrael783 Through an automation you can now toggle the silent mode the morning and the afternoon and reset it during the day.
You can also create a [script](https://www.home-assistant.io/integrations/script/#passing-variables-to-scripts) taking the entity id of your cover as input. And then use this script in a template for instance.

```yaml
script:
  open_slowly:
    fields:
      entity:
        description: "The cover to open slowly"
        example: "cover.bedroom"
    sequence:
      - service: switch.turn_on
        entity_id: switch.bedroom_low_speed
      - service: cover.open
        data:
          entity_id: "{{ entity }}"
      - service: switch.turn_off
        entity_id: switch.bedroom_low_speed
```

But to be checked, it seems you can call several services in a cover template: https://www.home-assistant.io/integrations/cover.template/#open_cover

iMicknl:
I think it will work fine, the only complexity is that you would need to capture the current status of the switch and restore the original status afterwards. (if you want to have it very clean).

However, I believe all solutions above are fine and we can just continue with this PR :-).

tetienne:
@azrael783 Can you confirm the switch are well created with your setup? Another user reported me it was not the case for him.

azrael783:
@tetienne I will check it this evening ;)

custom_components/tahoma/__init__.py
```diff
@@ -174,6 +175,9 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
         if device.widget == HOMEKIT_STACK:
             print_homekit_setup_code(device)
 
+    supported_platforms = set(platforms.keys())
+    # Required to load the low speed switch
+    supported_platforms.add(SWITCH)
```
tetienne, line 180: @iMicknl At the end we will have SWITCH, SENSOR, BINARY SENSOR. I guess we should add all the other ones (not in this PR). What do you think?

custom_components/tahoma/switch.py
tetienne: Done.

custom_components/tahoma/cover_devices/vertical_cover.py
```diff
@@ -101,16 +106,37 @@ def current_cover_position(self):
...
+                COMMAND_SET_CLOSURE_AND_LINEAR_SPEED, 0, "lowspeed"
+            )
+        else:
+            await self.executor.async_execute_command(
+                self.executor.select_command(*COMMANDS_OPEN)
+            )
 
     async def async_close_cover(self, **_):
         """Close the cover."""
-        await self.executor.async_execute_command(
```
tetienne, line 114: Fix current issue in master.

tetienne: @iMicknl

custom_components/tahoma/__init__.py
```diff
@@ -208,9 +209,8 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     supported_platforms = set(platforms.keys())
 
     # Sensor and Binary Sensor will be added dynamically, based on the device states
-    supported_platforms.add(BINARY_SENSOR)
-    supported_platforms.add(SENSOR)
-
+    # Required to load the low speed switch
```
iMicknl, line 212:
```suggestion
    # Switch will be added dynamically, based on device features (e.g. low speed cover switch)
```

Should we change it to something more descriptive?

custom_components/tahoma/cover_devices/tahoma_cover.py
```diff
@@ -92,6 +96,10 @@ def current_cover_tilt_position(self):
 
     async def async_set_cover_position_low_speed(self, **kwargs):
         """Move the cover to a specific position with a low speed."""
+        _LOGGER.warning(
+            "This service will be removed in Setptember 2021. Use instead the low speed switch."
```
iMicknl, line 100: Let's remove the warning and keep the service as well. At the core PR we can figure out if people miss this :-).

custom_components/tahoma/cover_devices/vertical_cover.py
```diff
@@ -101,16 +106,37 @@ def current_cover_position(self):
...
-            self.executor.select_command(*COMMANDS_CLOSE)
-        )
+        if self.is_low_speed_enabled():
```
iMicknl, line 129: Why not check first if this is a low speed cover? Doesn't make sense to gather the entity on every call. (even for non low speed covers)

tetienne: I will update `is_low_speed_enabled` with a safeguard avoid the entity lookup.

iMicknl (approved):
If tested and if it doesn't break other cover implementations, happy to merge :)

iMicknl:
Is this one ready to merge? Perhaps we can include this + DHCP + rebrand to Overkiz and make it 2.7.0.

tetienne:
Yes it should be OK. Sorry this weekend I was quite busy.

PR #498, opened 2021-07-28
Start rebranding component

tetienne:
The name of the component can confuse some people. Indeed, they stop reading at its name. "Oh I don’t have any Tahoma hub, that’s not for me."
Like already discussed  within a Core PR, I suggest *Overkiz (by Somfy)* and only rebrand the README.md for now, or perhaps do you want already change the name displayed in HACS (not the integration name).

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,15 +3,16 @@
 [![GitHub release](https://img.shields.io/github/release/iMicknl/ha-tahoma.svg)](https://GitHub.com/iMicknl/ha-tahoma/releases/)
 [![Open in Visual Studio Code](https://open.vscode.dev/badges/open-in-vscode.svg)](https://open.vscode.dev/iMicknl/ha-tahoma/)
 
-# Somfy TaHoma - Home Assistant
+# Overkiz (by Somfy) - Home Assistant
 
-Custom component for Home Assistant to interact with smart devices via the Somfy TaHoma platform. Despite the name of this integration, many other platforms sharing the same OverKiz API structure are supported as well. Have a look at all [supported hubs](#supported-hubs).
+Custom component for Home Assistant to interact with smart devices via the Overkiz platform.
 
 >This component builds upon the work of [@philklei](https://github.com/philklei) and is an updated version of the [original Tahoma integration](https://www.home-assistant.io/integrations/tahoma/) in Home Assistant with the goal of eventually merging into core. The installation of this component will replace the original TaHoma integration and thus allows you to beta-test [all changes](https://github.com/iMicknl/ha-tahoma/releases).
 
 ## Supported hubs
 
-- Somfy TaHoma Box
+- Somfy TaHoma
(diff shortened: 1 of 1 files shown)


vlebourl:
Is there any way to have aliases so that a google search of `Home Assistant Rexel` would lead straight to here? I don't think that Overkiz is very well known, although it's the common ground of all those devices...

tetienne:
What’s Rexel, that’s also a supported hub? I agree Overkiz is not well known. I don’t know what could be the best for our visibility.

vlebourl:
> What’s Rexel, that’s also a supported hub? 

I took the least obvious name in our Readme list of supported devices 😝

tetienne:
Damned, I didn’t scrool enough to see it. We can add the name of the hub as tag to our repo. It should help I guess.

README.md
```diff
@@ -3,15 +3,16 @@
 [![GitHub release](https://img.shields.io/github/release/iMicknl/ha-tahoma.svg)](https://GitHub.com/iMicknl/ha-tahoma/releases/)
 [![Open in Visual Studio Code](https://open.vscode.dev/badges/open-in-vscode.svg)](https://open.vscode.dev/iMicknl/ha-tahoma/)
 
-# Somfy TaHoma - Home Assistant
+# Overkiz (by Somfy) - Home Assistant
```
vlebourl, line 6: Maybe we could remove the parentheses?

iMicknl: What would be the benefit? In the core integration, we call everything Overkiz and thus it makes sense to rebrand here as well. For the core integration, auto discovery will help us enable every Overkiz hub user :-).

iMicknl:
> What’s Rexel, that’s also a supported hub? I agree Overkiz is not well known. I don’t know what could be the best for our visibility.

And there are even more... See https://github.com/iMicknl/ha-tahoma/pull/418/files. I would love to rebrand this whole integration to ha-overkiz as a breaking change and also implement changes like that one.

README.md
tetienne: @vlebourl You mean `Overkiz` or `Overkiz by Somfy`?

vlebourl:
I'm okay with it then, provided we use `Overkiz by Somfy` as a title.

README.md
vlebourl: I just thought `Overkiz by Somfy` is better than `Overkiz (by Somfy)`. But I won't fight for it, I'll accept the majority's opinion.

vlebourl (approved):

README.md
iMicknl: I can settle with both, so @tetienne you can make the final decision. 😉. I just wanted to not have Somfy to prominent visible, since it can scare users that their hubs won't be supported.

tetienne: That’s why I wrote it between braces to make it less "mandatory". We can also ask to the community on the forum. What do yo think?

PR #505, opened 2021-08-04
Add binary sensors dynamically based on available device state

iMicknl:
Similar to #425. This will allow us to add binary sensor easier and even multiple per device, where required.


tetienne:
I'm sorry I haven't any of these devices too. Perhaps you look within old issue if one on them was reported by an end user and ping him or try on the HA forum...

custom_components/tahoma/__init__.py
```diff
@@ -174,7 +176,12 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
         if device.widget == HOMEKIT_STACK:
             print_homekit_setup_code(device)
 
-    for platform in platforms:
+    supported_platforms = set(platforms.keys())
+
+    # Sensor and Binary Sensor will be added dynamic, based on the device states
```
tetienne, line 181: dynamic and dynamically have the same meaning?
```suggestion
    # Sensor and Binary Sensor will be added dynamically, based on the device states
```

custom_components/tahoma/binary_sensor.py
```diff
@@ -1,64 +1,101 @@
...
+@dataclass
+class OverkizBinarySensorDescription(BinarySensorEntityDescription):
+    """Class to describe a Overkiz binary sensor."""
+
+    is_on: Callable[[Any], Any] = lambda state: state
+
+
+BINARY_SENSOR_DESCRIPTIONS = [
+    # RainSensor/RainSensor
+    OverkizBinarySensorDescription(
```
tetienne, line 35: You removed 13 constants, but only added 9 descriptions, it’s on purpose?

iMicknl: Yes, I removed `core:ThreeWayHandleDirectionState` since this is a property that exists when other supported properties are available as well. I don't want to add multiple sensors for the same devices, that output the same info.

`core:ButtonState` since it is probably not used and an edge case that doesn't belong in `binary_sensor`.

However, it seems I missed `CORE_OPENING_STATE = "core:OpeningState"` and `CORE_OPEN_CLOSED_TILT_STATE = "core:OpenClosedTiltState"`. Will add them soon.

custom_components/tahoma/binary_sensor.py
```diff
@@ -1,64 +1,113 @@
...
+    # AirSensor/AirFlowSensor
+    OverkizBinarySensorDescription(
+        key="core:GasDetectionState",
+        name="Gas",
+        device_class=binary_sensor.DEVICE_CLASS_GAS,
+        is_on=lambda state: state == STATE_DETECTED,
+    ),
+    # WindowHandle/ThreeWayWindowHandle
+    OverkizBinarySensorDescription(
+        key="core:OpenClosedState",
```
iMicknl, line 64: This one triggers for my cover as well.. Should we keep it and add a binary sensor next to the cover implementation? 

Screen Shot 2021-08-09 at 17 05 07

iMicknl: @tetienne

iMicknl: I would maybe be in favor of removing `core:OpenClosedTiltState`, `core:OpenClosedState` and `core:OpeningState` since this should be part of cover already. My pergola also has `core:SlatsOpenClosedState` and we don't include this as well.

tetienne: That's OK for me. It will avoid to spam the device with too much entity.

iMicknl: Great! And if people miss it and report it, we can always see what to do :-).

custom_components/tahoma/binary_sensor.py
```diff
@@ -1,64 +1,94 @@
...
-    "WaterSensor": DEVICE_CLASS_WATER,
-    "WindowHandle": DEVICE_CLASS_OPENING,
-}
+
+
+@dataclass
+class OverkizBinarySensorDescription(BinarySensorEntityDescription):
+    """Class to describe a Overkiz binary sensor."""
+
+    is_on: Callable[[Any], Any] = lambda state: state
```
tetienne, line 30: State is already a boolean?

custom_components/tahoma/binary_sensor.py
```diff
@@ -69,55 +99,43 @@ async def async_setup_entry(
...
-        """Return the icon to use in the frontend, if any."""
-        if self.device_class == DEVICE_CLASS_WATER:
-            if self.is_on:
-                return ICON_WATER
-            return ICON_WATER_OFF
-
-        icons = {DEVICE_CLASS_GAS: ICON_WAVES, DEVICE_CLASS_RAIN: ICON_WEATHER_RAINY}
-
-        return icons.get(self.device_class)
+        state = self.select_state(self.entity_description.key)
```
tetienne, line 140: In such case, what do you think calling directly the get state from pyhoma? I would like to see select_state usage be reduced.

tetienne (approved):

custom_components/tahoma/binary_sensor.py
iMicknl: What do you mean? We use `is_on` to do the transformation, where we check e.g. is the state 'open' (binary = on), otherwise binary = off.

custom_components/tahoma/binary_sensor.py
iMicknl: Yes, makes sense. Will do this the same as the sensor.

custom_components/tahoma/binary_sensor.py
tetienne: Sorry, in my mind, it should returned a boolean because of the suffix `is_`.

iMicknl: The lambda will return a boolean? Could we type this in a better way?

The `is_on` is typed as Callable, since this is the lambda function that eventually will return a boolean.

PR #506, opened 2021-08-04
Split entity between entity and executor logic [refactor]

iMicknl:
Sorry in advance, this PR will be huge to review. I am trying to pull back the changes we had to make for core, in our custom component.

Core reviewers gave feedback that we should put the 'non entity' logic in a separate executor.. See https://github.com/home-assistant/core/tree/rework-tahoma/homeassistant/components/overkiz.

Already starting to rename some parts to Overkiz.


tetienne (dismissed):
It indeed makes entity well more clear and readable.

PR #512, opened 2021-08-10
Improve code quality for description entity

iMicknl:


custom_components/tahoma/entity.py
```diff
@@ -136,3 +141,43 @@ def device_state_attributes(self) -> Dict[str, Any]:
...
+    def name(self) -> str:
+        """Return the name of the device."""
+        if self.executor.index:
+            return f"{self.entity_description.name} {self.executor.index}"
+        return self.entity_description.name
+
+    @property
+    def unique_id(self) -> str:
+        """Return a unique ID."""
+        return f"{super().unique_id}-{self.entity_description.key}"
```
iMicknl, line 183: @tetienne my only concern is this one, do we need to add the platform here as well? (thus moving unique id to binary_sensory.py and sensor.py)?

I don't think that it wil happen that a sensor and binary sensor share the same state, but maybe I am overlooking anything?

custom_components/tahoma/entity.py
```diff
@@ -136,3 +141,43 @@ def device_state_attributes(self) -> Dict[str, Any]:
...
+    """Class to describe an Overkiz sensor."""
+
+    value: Callable[[str | int | float], str | int | float] | None = lambda val: val
+
+
+@dataclass
+class OverkizBinarySensorDescription(BinarySensorEntityDescription):
+    """Class to describe an Overkiz binary sensor."""
+
+    is_on: Callable[[str], bool] = lambda state: state
```
tetienne, line 157:
```suggestion
    is_on: Callable[str, bool] = lambda state: state
```

custom_components/tahoma/entity.py
tetienne: I think we are safe. The state name is unique among a device.

tetienne (approved):

PR #516, opened 2021-08-11
Clean base url an index

### LINKED ISSUE #513: RSSI level sensors naming

### The problem

With the most recent version every RSSI sensor is named RSSi_level_x
It's impossible to tell which is which. 

Screenshot_20210810-200128_Samsung Internet

### What version of this integration (ha-tahoma) has the issue?

2.6.0

### What version of Home Assistant Core has the issue?

Core-2021.8.5

Tahoma V2 Hub

### Device model

Positionable Roller shutter with low speed

### Device type

Shutter b

### Additional information

_No response_

### Relevant log output

_No response_

tetienne:
Fix #513

### DIFF
--- a/custom_components/tahoma/entity.py
+++ b/custom_components/tahoma/entity.py
@@ -173,9 +173,7 @@ class OverkizDescriptiveEntity(OverkizEntity):
     @property
     def name(self) -> str:
         """Return the name of the device."""
-        if self.executor.index:
-            return f"{self.entity_description.name} {self.executor.index}"
-        return self.entity_description.name
+        return f"{super().name} {self.entity_description.name}"
 
     @property
     def unique_id(self) -> str:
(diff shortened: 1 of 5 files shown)


custom_components/tahoma/entity.py
```diff
@@ -54,6 +54,7 @@ def __init__(self, device_url: str, coordinator: TahomaDataUpdateCoordinator):
         """Initialize the device."""
         super().__init__(coordinator)
         self.device_url = device_url
+        self.base_device_url, *_ = self.device_url.split("#")
```
iMicknl, line 57: Isn't this entry already present in the executor? self.executor.base_device_url?

tetienne: Yes indeed. But it sounds weird to retrieve the executor to get this data while device_url is directly accessible.

iMicknl (approved):

Kaibob2:
Just installed 2.6.1 and can confirm, that the RSSI sensors now follow the convention. Thanks for that.

PR #520, opened 2021-08-11
Add DHCP discovery to Config Flow

iMicknl:
Ported from our core PR. I didn't know that custom components support this feature as well.


iMicknl:
Not sure why tests fail, locally they run without issues..

tetienne (approved):

Issue #476, opened 2021-07-15
Unexpected error fetching device events data: Too many requests, try again later : login with email - custom_components/tahoma/coordinator.py:78


Wiggum127:
### Did you read the instructions?

- [X] I have read the [README](https://github.com/iMicknl/ha-tahoma/blob/master/README.md), including the [advanced debugging](https://github.com/iMicknl/ha-tahoma/blob/master/README.md#advanced) section.

### The problem

Getting many errors lately which I didn't have before.
The logging claims there have been too many login attempts, for some reason.

Logging on to the Tahoma platform via the Somfy app is not an issue. The user account/credentials are ok.

I've tried lowering the polling interval from the default 30s to 300s and disabled the integration for a while.
After enabling the integration again, I get the same errors back.

An automation I have made for lowering screens at the end of the day no longer works. Interactively lowering a screen works.

### What version of this integration (ha-tahoma) has the issue?

2.5.0

### What version of Home Assistant Core has the issue?

core-2021.7.2

### Which gateway / hub do you use?

Somfy Tahoma

### Device model

PositionableScreen

### Device type

io:VerticalExteriorAwningIOComponent

### Additional information
```
Logger: custom_components.tahoma
Source: custom_components/tahoma/coordinator.py:78
Integration: Somfy TaHoma
First occurred: 27 juni 2021 14:09:22 (471 occurrences)
Last logged: 14:44:50

Unexpected error fetching device events data: Too many requests, try again later : login with email@e.mail
Traceback (most recent call last):
  File "/config/custom_components/tahoma/coordinator.py", line 67, in _async_update_data
    events = await self.client.fetch_events()
  File "/usr/local/lib/python3.8/site-packages/pyhoma/client.py", line 215, in fetch_events
    response = await self.__post(f"events/{self.event_listener_id}/fetch")
  File "/usr/local/lib/python3.8/site-packages/pyhoma/client.py", line 333, in __post
    await self.check_response(response)
  File "/usr/local/lib/python3.8/site-packages/pyhoma/client.py", line 371, in check_response
    raise NotAuthenticatedException(message)
pyhoma.exceptions.NotAuthenticatedException: Not authenticated

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/update_coordinator.py", line 187, in _async_refresh
    self.data = await self._async_update_data()
[...]

PR #524, opened 2021-08-12
Improve error handling in coordinator (during BadCredentials error)

iMicknl:


custom_components/tahoma/coordinator.py
```diff
@@ -76,8 +76,13 @@ async def _async_update_data(self) -> Dict[str, Device]:
             raise UpdateFailed("Failed to connect.") from exception
         except (ServerDisconnectedError, NotAuthenticatedException):
             self.executions = {}
-            await self.client.login()
-            self.devices = await self._get_devices()
+
+            try:
+                await self.client.login()
+                self.devices = await self._get_devices()
+            except BadCredentialsException as exception:
```
tetienne, line 83: How it can be possible?

iMicknl: I have seen this more often, however it happens in the following way:

1. Somfy has some kind of error and logs us out
2. We get a `NotAuthenticatedException` and will do a new login
3. New login will get a BadCredentialsException due to Somfy server issue (and retry a few times due to backoff)
4. We are blacklisted temporarily and throw a BadCredentialsException again here, that we don't catch
5. We keep retrying and the temporary ban will extend every time.

custom_components/tahoma/coordinator.py
```diff
@@ -76,8 +76,16 @@ async def _async_update_data(self) -> Dict[str, Device]:
...
-            await self.client.login()
-            self.devices = await self._get_devices()
+
+            # During the relogin, similar exceptions can be thrown.
+            try:
+                await self.client.login()
+                self.devices = await self._get_devices()
+            except BadCredentialsException as exception:
+                raise ConfigEntryAuthFailed() from exception
+            except TooManyRequestsException as exception:
```
iMicknl, line 86: Added `TooManyRequestsException` as well, to have better visual errors for https://github.com/iMicknl/ha-tahoma/issues/476.

tetienne (approved):

Issue #503, opened 2021-08-01
Convert @property to _attr_variable


iMicknl:
### Did you read the instructions?

- [X] I have read the [README](https://github.com/iMicknl/ha-tahoma/blob/master/README.md), including the [advanced debugging](https://github.com/iMicknl/ha-tahoma/blob/master/README.md#advanced) section.

### The request

In core, they made a change to add attr_ variables instead of setting properties. It would be good to see if we can make this change as well.

See https://github.com/home-assistant/core/pull/53816 for an example. See https://developers.home-assistant.io/docs/core/entity/#entity-class-or-instance-attributes for more context.

PR #532, opened 2021-08-16
Use instance variable for name

tetienne:
Part of #503


iMicknl (approved):
LGTM, assume it's tested and doesn't have any issues? 😉

tetienne:
> LGTM, assume it's tested and doesn't have any issues? 😉

Sure!

Issue #279, opened 2020-09-29
Newly added devices require a restart of home assistant


golles:
- [x] **I have read the [Readme](https://github.com/iMicknl/ha-tahoma/blob/master/README.md), including the [Advanced](https://github.com/iMicknl/ha-tahoma/blob/master/README.md#advanced) section regarding debugging.**

**Describe the bug**
Today my 2 new roller shutters (IO) were installed and I've added them in the Somfy app. I had to restart HA to make them appear there. I would expect that a restart would not be needed (in the final version of this component)

**To Reproduce**
Add new devices to Somfy app/website (https://www.tahomalink.com/)
Devices not appearing in HA
following logs are appearing:
`2020-09-29 18:37:00 DEBUG (MainThread) [custom_components.tahoma] Finished fetching TaHoma Event Fetcher data in 0.108 seconds`
Devices not appearing in HA after waiting for 10-15 minutes
Reboot HA
Devices are there :-)

**Expected behavior**
Newly added devices get added to HA without restarting

**Screenshots**
None

**Environment (please complete the following information):**

- **ha-tahoma version:** master, commit: 7c1f11d2158dfeec74ebd6434fdade89128041b5 (latest available as of creating this ticket)
- **Home Assistant version:** 0.115.5, also latest
- **Platform**: cover

**Device:**
Roller shutter

- **Model**: PositionableRollerShutter
- **Type**: io:RollerShutterGenericIOComponent

**Additional context**
I guess the component is not mature enough to support this, maybe it's not a bug but a feature request. 
Again I want to express that you're all doing an amazing job making this new component, I'm very happy with it!

iMicknl:
Thanks for reporting @golles! This is something that we haven't implemented (yet) indeed. We already catch the `DeviceCreatedEvent`, however we don't reload the integration at the moment.

This is something that has a bit of a lower prio than the other open issues, since you can currently 'reload' the integration via the UI, which will add them as well. A full restart is not needed anymore since 0.115.

https://github.com/iMicknl/ha-tahoma/blob/7c1f11d2158dfeec74ebd6434fdade89128041b5/custom_components/tahoma/coordinator.py#L79-L83

golles:
I'm not planning to buy any new devices soon, so won't be able to test any time soon :-)

toby200:
[...]

PR #533, opened 2021-08-16
Handle device created / updated / removed event in DataUpdateCoordinator

iMicknl:
Implements #279

### DIFF
--- a/custom_components/tahoma/coordinator.py
+++ b/custom_components/tahoma/coordinator.py
@@ -19,6 +19,8 @@ from pyhoma.exceptions import (
 )
 from pyhoma.models import DataType, Device, Place, State
 
+from .const import DOMAIN
+
 TYPES = {
     DataType.NONE: None,
     DataType.INTEGER: int,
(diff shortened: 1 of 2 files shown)


custom_components/tahoma/coordinator.py
```diff
@@ -114,11 +118,18 @@ async def _async_update_data(self) -> Dict[str, Device]:
                 EventName.DEVICE_CREATED,
                 EventName.DEVICE_UPDATED,
             ]:
-                self.devices = await self._get_devices()
+                self.hass.async_create_task(
+                    self.hass.config_entries.async_reload(self._config_entry_id)
+                )
+                return None
```
tetienne, line 124: Not sure about this one. Why return  None?

custom_components/tahoma/coordinator.py
```diff
@@ -114,11 +118,18 @@ async def _async_update_data(self) -> Dict[str, Device]:
...
 
             elif event.name == EventName.DEVICE_REMOVED:
+                base_device_url, *_ = event.deviceurl.split("#")
                 registry = await device_registry.async_get_registry(self.hass)
-                registry.async_remove_device(event.deviceurl)
+
+                if device := registry.async_get_device({(DOMAIN, base_device_url)}):
```
tetienne, line 130: 1st time I see someone use this :)

iMicknl: I took it from here https://github.com/custom-components/integration_blueprint/pull/60/files :D

iMicknl: I took it from uptimerobot in core! https://github.com/home-assistant/core/blob/8103d9ae3c12664cdec27d742481cca28eee0796/homeassistant/components/uptimerobot/__init__.py#L79

custom_components/tahoma/coordinator.py
iMicknl: I am not 100% sure, I took it from core: https://github.com/home-assistant/core/blob/8103d9ae3c12664cdec27d742481cca28eee0796/homeassistant/components/uptimerobot/__init__.py#L79.

However my belief was that we ask the integration to reload and don't return anything in this update cycle, since it doesn't make sense. The integration will reload and refresh everything anyway.

tetienne (approved):

PR #540, opened 2021-08-19
Clean up entity model, remove extra states and attributes

iMicknl:
- Remove extra states and attributes
- Improve device info typing
- Use extra_state_attributes (new in core)
- Only use the actual firmware for firmware

### DIFF
--- a/custom_components/tahoma/entity.py
+++ b/custom_components/tahoma/entity.py
@@ -18,6 +18,7 @@ ATTR_RSSI_LEVEL = "rssi_level"
 
 CORE_AVAILABILITY_STATE = "core:AvailabilityState"
 CORE_BATTERY_STATE = "core:BatteryState"
+CORE_FIRMWARE_REVISION = "core:FirmwareRevision"
 CORE_MANUFACTURER = "core:Manufacturer"
 CORE_MANUFACTURER_NAME_STATE = "core:ManufacturerNameState"
 CORE_MODEL_STATE = "core:ModelState"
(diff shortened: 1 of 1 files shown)


custom_components/tahoma/entity.py
```diff
@@ -88,40 +89,28 @@ def device_info(self) -> dict[str, Any]:
             or self.device.widget
         )
 
-        return {
-            "identifiers": {(DOMAIN, self.executor.base_device_url)},
-            "manufacturer": manufacturer,
-            "name": self.device.label,
-            "model": model,
-            "sw_version": self.device.controllable_name,
```
tetienne, line 96: I would like to see the device.controllable_name somewhere. It helps me do retrieve the setup.

iMicknl: Same! I would propose to add this to system_health and update our README? Or only keep this in the debug logging and ask people to add their debug logs.

It doesn't make a lot of sense to have this visible in the main interface, if we only need it for debug purposes. Agree?

tetienne: Agree! Can you add the debug log with this PR?

iMicknl: @tetienne don't we already have all this information in the debug log? Only system health isn't in this PR yet.

tetienne (approved):

tetienne (approved):

tetienne:
Can we merge this one?

PR #542, opened 2021-08-19
Add sensor for Discrete RSSI Level (disabled by default)

iMicknl:
Based on feedback on the community forum, disabled by default since I don't see the value for all users.

Screen Shot 2021-08-19 at 16 41 21

### DIFF
--- a/custom_components/tahoma/sensor.py
+++ b/custom_components/tahoma/sensor.py
@@ -307,6 +307,13 @@ SENSOR_DESCRIPTIONS = [
         icon="mdi:content-save-cog",
         entity_registry_enabled_default=False,
     ),
+    OverkizSensorDescription(
+        key="core:DiscreteRSSILevelState",
+        name="Discrete RSSI Level",
+        entity_registry_enabled_default=False,
+        native_value=lambda value: str(value).capitalize(),
+        device_class=sensor.DEVICE_CLASS_SIGNAL_STRENGTH,
+    ),
 ]
 
 


custom_components/tahoma/sensor.py
```diff
@@ -289,6 +298,20 @@
...
+        key="core:Memorized1PositionState",
+        name="Memorized Position",
+        entity_registry_enabled_default=False,
+        icon="mdi:content-save-cog",
+    ),
+    OverkizSensorDescription(
+        key="core:DiscreteRSSILevelState",
+        name="Discrete RSSI Level",
+        entity_registry_enabled_default=False,
+        value=lambda value: str(value).capitalize(),
```
tetienne, line 312: Why capitalize ?

custom_components/tahoma/sensor.py
```diff
@@ -22,6 +22,14 @@
 from .const import DOMAIN
 from .entity import OverkizDescriptiveEntity, OverkizSensorDescription
 
+OVERKIZ_BATTERY_TO_HOME_ASSISTANT = {
```
tetienne, line 25: That’s a really long name :)

```suggestion
OVERKIZ_BATTERY_TO_HA = {
```

custom_components/tahoma/sensor.py
```diff
@@ -34,7 +42,8 @@
         key="core:BatteryState",
         name="Battery",
         device_class=sensor.DEVICE_CLASS_BATTERY,
-        value=lambda value: str(value).capitalize(),
+        unit_of_measurement=PERCENTAGE,
+        value=lambda value: OVERKIZ_BATTERY_TO_HOME_ASSISTANT.get(value, value),
```
tetienne, line 46:
```suggestion
        value=lambda value: OVERKIZ_BATTERY_TO_HA.get(value, value),
```

tetienne: Why default to value?

tetienne:
Defect sensor: 

image

👌

custom_components/tahoma/sensor.py
iMicknl: We do this for other sensors as well. We get ‘low’ as input and I thought that other integrations always capitalize the first letter in a text sensors, thus ‘Low’.

But we can also revert this change throughout the whole integration.

tetienne:
I thought my LightSensor has a battery state because Tahomalink show me a battery icon, but that’s not the case. 

image

In fact, they display this logo when there is no defect: `<div class="battery-level noDefect"></div>
`

custom_components/tahoma/binary_sensor.py
```diff
@@ -73,6 +73,12 @@
         device_class=binary_sensor.DEVICE_CLASS_VIBRATION,
         is_on=lambda state: state == STATE_DETECTED,
     ),
+    OverkizBinarySensorDescription(
```
tetienne, line 76: That’s much "complex" than this. There is 4 available values: 

```JSON
{
      "type": "DiscreteState",
      "values": [
        "dead",
        "lowBattery",
        "maintenanceRequired",
        "noDefect"
      ],
      "qualifiedName": "core:SensorDefectState"
    },
```

See what does the homebridge integration: https://github.com/dubocr/homebridge-tahoma/blob/97bd290ff9cb466ca6095c694518e9a931b93f60/src/mappers/RainSensor.ts#L24

custom_components/tahoma/sensor.py
iMicknl: This was the previous behavior and will catch a situation when Somfy (or another vendor) has a different key in the `core:BatteryState`.

custom_components/tahoma/binary_sensor.py
iMicknl: Not sure how to achieve this in our case... Where it is split between binary sensor and sensor.

tetienne: I think we really need to create a BatterySensor class to handle the logic. We can detect battery device though an attribute `core:PowerSourceType`:

```
"attributes": [
                {
                    "name": "core:MinSensedValue",
                    "type": 1,
                    "value": 50
                },
                {
                    "name": "core:MaxSensedValue",
                    "type": 1,
                    "value": 100000
                },
                {
                    "name": "core:MeasuredValueType",
                    "type": 3,
                    "value": "core:LuminanceInLux"
                },
                {
                    "name": "core:Manufacturer",
                    "type": 3,
                    "value": "Somfy"
                },
                {
                    "type": 3,
                    "value": "battery"
                }
            ],
```

iMicknl: Fine by me, but will still be pretty complex if we also want to handle situations like `maintenanceRequired`.

custom_components/tahoma/sensor.py
tetienne: That’s fine. I thought it was LOW.

custom_components/tahoma/binary_sensor.py
tetienne: As did homebridge, we can map maintenanceRequired to dead.

iMicknl: > As did homebridge, we can map maintenanceRequired to dead.

But currently we cannot create a binary_sensor and sensor for the same state key.

tetienne: So we just have to create a sensor and not a binary sensor.

iMicknl:
@tetienne moved our battery discussion to https://github.com/iMicknl/ha-tahoma/issues/580. Let's just get this PR in and than move the battery step to a new PR.

tetienne (approved):

PR #543, opened 2021-08-20
Add discord link

tetienne:


iMicknl:
Does this link have a long lifetime? (thus not a week :D)

And perhaps we can rename the server eventually to Overkiz (by Somfy) custom development or whatever.

iMicknl (approved):

tetienne:
@iMicknl The link use the server id, so yes that’s safe :)

PR #545, opened 2021-08-24
Add support for priority lock

tetienne:
Screenshot 2021-08-24 at 09-34-26 Configuration - Home Assistant
Screenshot 2021-08-24 at 09-33-02 Configuration - Home Assistant

### DIFF
--- a/custom_components/tahoma/cover_devices/tahoma_cover.py
+++ b/custom_components/tahoma/cover_devices/tahoma_cover.py
@@ -54,7 +54,6 @@ CORE_OPEN_CLOSED_STATE = "core:OpenClosedState"
 CORE_OPEN_CLOSED_UNKNOWN_STATE = "core:OpenClosedUnknownState"
 # io:DiscreteGateOpenerIOComponent
 CORE_PEDESTRIAN_POSITION_STATE = "core:PedestrianPositionState"
-CORE_PRIORITY_LOCK_TIMER_STATE = "core:PriorityLockTimerState"
 CORE_SLATS_OPEN_CLOSED_STATE = "core:SlatsOpenClosedState"
 CORE_SLATE_ORIENTATION_STATE = "core:SlateOrientationState"
 CORE_SLATS_ORIENTATION_STATE = "core:SlatsOrientationState"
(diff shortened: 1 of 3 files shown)


custom_components/tahoma/sensor.py
```diff
@@ -289,6 +285,21 @@
...
+    OverkizSensorDescription(
+        key="io:PriorityLockOriginatorState",
+        name="Priority Lock Originator",
+        value=lambda value: str(value).capitalize(),
+        icon="mdi:lock",
+        entity_registry_enabled_default=False,
+    ),
+    OverkizSensorDescription(
+        key="core:PriorityLockTimerState",
+        value=lambda value: value,
```
iMicknl, line 297: You can remove this, if you don't want to do any transformation.

custom_components/tahoma/sensor.py
```diff
@@ -289,6 +285,21 @@
         value=lambda value: str(value).capitalize(),
         entity_registry_enabled_default=False,
     ),
+    OverkizSensorDescription(
+        key="io:PriorityLockOriginatorState",
+        name="Priority Lock Originator",
+        value=lambda value: str(value).capitalize(),
```
iMicknl, line 291: What will be the default value, if there is nothing? Will str(value) be 'none'?

tetienne, line 291: I hesitate with *Unlocked* as default value for this one, instead of *Unknown*.

iMicknl: Nevermind, it seems that this is catched below :-)

tetienne: Look at my screenshot :)

tetienne: Look at my screenshot :)

iMicknl: Which screenshot? ;-)

tetienne: Damned, they were not updated. Should be ok now.

iMicknl (approved):

custom_components/tahoma/cover_devices/tahoma_cover.py
```diff
@@ -128,19 +123,6 @@ def is_closed(self):
 
         return None
 
-    @property
-    def icon(self):
```
iMicknl, line 132: Finally we got rid of this part <3

tetienne: 💪

PR #556, opened 2021-09-08
Add My Position sensor

tetienne:
This sensor is disabled by default.

### DIFF
--- a/custom_components/tahoma/cover_devices/tahoma_cover.py
+++ b/custom_components/tahoma/cover_devices/tahoma_cover.py
@@ -45,7 +45,6 @@ COMMANDS_SET_TILT_POSITION = [COMMAND_SET_ORIENTATION]
 CORE_CLOSURE_STATE = "core:ClosureState"
 CORE_CLOSURE_OR_ROCKER_POSITION_STATE = "core:ClosureOrRockerPositionState"
 CORE_DEPLOYMENT_STATE = "core:DeploymentState"
-CORE_MEMORIZED_1_POSITION_STATE = "core:Memorized1PositionState"
 CORE_MOVING_STATE = "core:MovingState"
 CORE_OPEN_CLOSED_PARTIAL_STATE = "core:OpenClosedPartialState"
 # io:DiscreteGateOpenerIOComponent
--- a/custom_components/tahoma/sensor.py
+++ b/custom_components/tahoma/sensor.py
@@ -303,7 +303,8 @@ SENSOR_DESCRIPTIONS = [
     ),
     OverkizSensorDescription(
         key="core:Memorized1PositionState",
-        name="MY position",
+        name="My Position",
+        icon="mdi:content-save-cog",
         entity_registry_enabled_default=False,
     ),
 ]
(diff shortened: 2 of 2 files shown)


custom_components/tahoma/cover_devices/tahoma_cover.py
```diff
@@ -45,7 +45,6 @@
 CORE_CLOSURE_STATE = "core:ClosureState"
 CORE_CLOSURE_OR_ROCKER_POSITION_STATE = "core:ClosureOrRockerPositionState"
 CORE_DEPLOYMENT_STATE = "core:DeploymentState"
-CORE_MEMORIZED_1_POSITION_STATE = "core:Memorized1PositionState"
```
tetienne, line 48: This constant was not used.

custom_components/tahoma/sensor.py
```diff
@@ -301,6 +301,11 @@
         native_unit_of_measurement=TIME_SECONDS,
         entity_registry_enabled_default=False,
     ),
+    OverkizSensorDescription(
+        key="core:Memorized1PositionState",
+        name="MY position",
```
iMicknl, line 306: Why not 'My Position'? Is MY how Somfy writes it in English, in Dutch they just use My.
And we capitalize all words in sensors for now, maybe we should do that differently, but would be good to have them all the same.

custom_components/tahoma/sensor.py
```diff
@@ -301,6 +301,11 @@
...
+        entity_registry_enabled_default=False,
```
iMicknl, line 307: Could we add a default icon? Like the one I added here: https://github.com/iMicknl/ha-tahoma/pull/542.

This PR also implemented the same sensor by the way.

Mariusthvdb: since you mention the #542, maybe differentiate there between the 2 sensors? check mdi for 'signal' and a few good candidates emerge, or use wifi-strength for either of the 2.

sorry if this is off-topic here, let me know if you need a dedicated FR/PR for that

iMicknl: I am not sure what you mean @Mariusthvdb, however for RSSI we don't set an icon. We set the         `device_class=sensor.DEVICE_CLASS_SIGNAL_STRENGTH` and front-end takes care of the icon.

custom_components/tahoma/sensor.py
tetienne: When I look for it, I always find it in upper case: https://service.somfy.com/downloads/fr_v4/programmation_des_fins_de_courses_vr.pdf

You suggest `My Position`?

custom_components/tahoma/sensor.py
tetienne: I thought about ⭐ like favorite, but also like you said about 💾 then I was confused and let the default one.

Mariusthvdb: > I am not sure what you mean @Mariusthvdb, however for RSSI we don't set an icon. We set the `device_class=sensor.DEVICE_CLASS_SIGNAL_STRENGTH` and front-end takes care of the icon.

I see. Its just that having the same icon for 2 different (discrete and numerical) sensors can be a bit visually confusing. Since they both pertain the RSII level, maybe not to worry.
and as before, an icon template is easily made.

iMicknl (approved):

PR #561, opened 2021-09-10
Remove polling custom interval and reduce it for stateless configuration

tetienne:
To integrate Home Assistant Core, we must remove this feature. This one will be refused by the Core team.

### DIFF
--- a/custom_components/tahoma/__init__.py
+++ b/custom_components/tahoma/__init__.py
@@ -1,7 +1,7 @@
 """The Overkiz (by Somfy) integration."""
 import asyncio
 from collections import defaultdict
-from datetime import timedelta
+from datetime import datetime
 from enum import Enum
 import logging
 
@@ -33,12 +33,11 @@ import voluptuous as vol
 
 from .const import (
     CONF_HUB,
-    CONF_UPDATE_INTERVAL,
     DEFAULT_HUB,
-    DEFAULT_UPDATE_INTERVAL,
     DOMAIN,
     IGNORED_OVERKIZ_DEVICES,
     OVERKIZ_DEVICE_TO_PLATFORM,
+    UPDATE_INTERVAL,
 )
 from .coordinator import OverkizDataUpdateCoordinator
 
@@ -157,10 +156,6 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
         _LOGGER.exception(exception)
         return False
 
-    update_interval = timedelta(
-        seconds=entry.options.get(CONF_UPDATE_INTERVAL, DEFAULT_UPDATE_INTERVAL)
-    )
-
     coordinator = OverkizDataUpdateCoordinator(
         hass,
         _LOGGER,
@@ -168,12 +163,8 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
         client=client,
         devices=devices,
         places=places,
-        update_interval=update_interval,
+        update_interval=UPDATE_INTERVAL,
         config_entry_id=entry.entry_id,
-    )
-
-    _LOGGER.debug(
-        "Initialized DataUpdateCoordinator with %s interval.", str(update_interval)
     )
 
     await coordinator.async_refresh()
@@ -184,7 +175,6 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     hass.data[DOMAIN][entry.entry_id] = {
         "platforms": platforms,
         "coordinator": coordinator,
-        "update_listener": entry.add_update_listener(update_listener),
     }
 
     # Map Overkiz device to Home Assistant platform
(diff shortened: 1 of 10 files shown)


iMicknl:
@tetienne would you perhaps be able to tackle the RTS issue as well? It would be good to have a different timeout (by default) for RTS only installations, and don't do the 1 second polling for executions of RTS devices.

tetienne:
OK it makes sense. I will have a look too.

vlebourl (approved):

vlebourl (commented):
I approved maybe too quickly, are we sure 30s is good enough for every device?

tetienne:
@vlebourl what do you mean? Somfy disapproved having a polling interval lower than 30s.

custom_components/tahoma/coordinator.py
```diff
@@ -146,7 +147,7 @@ async def _async_update_data(self) -> Dict[str, Device]:
                 del self.executions[event.exec_id]
 
         if not self.executions:
```
tetienne, line 149: @iMicknl IMO here can be the issue for the too many request. If we missed an event, the executions list will be never empty.

iMicknl: Let's have a better look. Maybe we should make sure to reset the update interval after we have to refresh all devices. I thought this logic was somewhere, but I guess this PR never got merged in.

iMicknl (commented):
Nice! Looks good. I will have a better look later.

custom_components/tahoma/coordinator.py
```diff
@@ -59,9 +59,9 @@ def __init__(
         )
 
         self.data = {}
-        self.original_update_interval = update_interval
         self.client = client
         self.devices: Dict[str, Device] = {d.deviceurl: d for d in devices}
+        self.is_stateless = all(len(device.states) == 0 for device in devices)
```
tetienne, line 64: For this, I don’t know if it can happen. For instante, just the Tahoma device has 3 states: firmware, Manufacturer, and SupportedManufacturerSettingsCommands

tetienne: I should perhaps add a condition on the type ACTUACTOR. But there is also sensor. What do you thing?

iMicknl: Are you sure that firmware and manufacturer is a state? Don't they put that in the attributes?

tetienne: Hum you are right, I read to quickly.

iMicknl: I don't own any RTS devices, are we sure that RTS devices (and hubs) don't have anything in the states?

tetienne: I have no clue too :(

iMicknl: Let's just merge it, it cannot be worse than before. And based on user complaints / logging, we can investigate further.

custom_components/tahoma/__init__.py
```diff
@@ -157,34 +156,32 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
...
 
-    _LOGGER.debug(
-        "Initialized DataUpdateCoordinator with %s interval.", str(update_interval)
-    )
+    await coordinator.async_config_entry_first_refresh()
 
-    await coordinator.async_refresh()
+    if coordinator.is_stateless:
+        _LOGGER.debug(
+            "All devices have assumed state. Update interval has been reduced to: %s",
```
iMicknl, line 174:
```suggestion
            "All devices have an assumed state. Update interval has been reduced to: %s",
```

```suggestion
            "All devices have are stateless. Update interval has been reduced to: %s",
```

tetienne: I took it from bdraco on the Somfy component :P

custom_components/tahoma/coordinator.py
```diff
@@ -146,7 +147,7 @@ async def _async_update_data(self) -> Dict[str, Device]:
                 del self.executions[event.exec_id]
 
         if not self.executions:
-            self.update_interval = self.original_update_interval
+            self.update_interval = UPDATE_INTERVAL
```
iMicknl, line 150: Does this take RTS into account?

tetienne: For RTS, we don’t update the interval, so that’s ok.

custom_components/tahoma/__init__.py
```diff
@@ -33,12 +32,12 @@
...
 from .const import (
     CONF_HUB,
-    CONF_UPDATE_INTERVAL,
     DEFAULT_HUB,
-    DEFAULT_UPDATE_INTERVAL,
     DOMAIN,
     IGNORED_OVERKIZ_DEVICES,
     OVERKIZ_DEVICE_TO_PLATFORM,
+    UPDATE_INTERVAL,
+    UPDATE_INTERVAL_ALL_ASSUMED_STATE,
```
iMicknl, line 40:
```suggestion
    UPDATE_INTERVAL_RTS, # RTS devices have an assumed state only
```

Should we just call it RTS to keep it more Somfy oriented? I understand RTS easier than assumed state. And we can add a comment for other maintainers / core contributors to understand the type.

iMicknl (dismissed):

custom_components/tahoma/coordinator.py
```diff
@@ -59,9 +59,13 @@ def __init__(
         )
 
         self.data = {}
-        self.original_update_interval = update_interval
         self.client = client
         self.devices: Dict[str, Device] = {d.deviceurl: d for d in devices}
+        self.is_stateless = all(
+            device.deviceurl.startswith("rts://")
```
tetienne, line 65: Would be nice to add a protocol property within pyhoma

iMicknl: An enum? Or a property in the model? 

Could we perhaps merge this one in for the mean-time? It should fix quite some users problems.

tetienne: Just a property. OK to be merged if has been tested

iMicknl: I tested it and it will change the update interval for RTS installations, and keep the same one for non RTS.

iMicknl (approved):

PR #570, opened 2021-09-20
Update codeowners

tetienne:
Even if we fork philklei initial work, now our custom component is totally different and does not behave like the legacy one. Moreover, philklei was never involved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 
 Custom component for Home Assistant to interact with smart devices via the Overkiz platform.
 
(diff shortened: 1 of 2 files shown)


tetienne:
@iMicknl A quick approval?

iMicknl:
> @iMicknl A quick approval?

👍🏼. Do you want to change the README as well?

iMicknl (approved):

tetienne:
@iMicknl Good catch, I missed this part.

iMicknl (approved):

PR #573, opened 2021-09-21
Trigger event on failure

tetienne:
Some events are not persisted though states. For this reason, and to allow automation, we can fire our own event. 

```JSON
{
    "event_type": "overkiz.event",
    "data": {
        "event_name": "ExecutionStateChangedEvent",
        "failure_type_code": 106,
        "failure_type": "CMDCANCELLED"
    },
    "origin": "LOCAL",
    "time_fired": "2021-09-28T20:03:57.102478+00:00",
    "context": {
        "id": "92a84240d914b43ceaf1aee3249568f6",
        "parent_id": null,
        "user_id": null
    }
}
```

TODO:
- [x] documentation

### DIFF
--- a/custom_components/tahoma/coordinator.py
+++ b/custom_components/tahoma/coordinator.py
@@ -98,6 +98,22 @@ class OverkizDataUpdateCoordinator(DataUpdateCoordinator):
         for event in events:
             _LOGGER.debug(event)
 
+            if event.failure_type_code:
+                # Some event names are not mapped to an enum, it’s why this trick.
+                # Like CommandExecutionStateChangedEvent
+                try:
+                    raw_event_name = event.name.value
+                except AttributeError:
+                    raw_event_name = event.name
+
+                self.hass.bus.fire(
+                    raw_event_name,
+                    {
+                        "failure_type_code": event.failure_type_code.value,
+                        "failure_type": event.failure_type,
+                    },
+                )
+
             if event.name == EventName.DEVICE_AVAILABLE:
                 self.devices[event.deviceurl].available = True
 
--- a/custom_components/tahoma/coordinator.py
+++ b/custom_components/tahoma/coordinator.py
@@ -99,12 +99,7 @@ class OverkizDataUpdateCoordinator(DataUpdateCoordinator):
             _LOGGER.debug(event)
 
             if event.failure_type_code:
-                # Some event names are not mapped to an enum, it’s why this trick.
-                # Like CommandExecutionStateChangedEvent
-                try:
-                    raw_event_name = event.name.value
-                except AttributeError:
-                    raw_event_name = event.name
+                raw_event_name = event.name.value
 
                 self.hass.bus.fire(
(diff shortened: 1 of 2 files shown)


custom_components/tahoma/coordinator.py
```diff
@@ -105,6 +105,22 @@ async def _async_update_data(self) -> Dict[str, Device]:
                 event.new_state,
             )
 
+            if event.failure_type_code:
```
tetienne, line 108: Wondering if cannot make this generics, instead of limiting this to the failure.

tetienne, line 108: Wondering if cannot make this generics, instead of limiting this to the failure.

custom_components/tahoma/coordinator.py
```diff
@@ -105,6 +105,22 @@ async def _async_update_data(self) -> Dict[str, Device]:
...
+                # Some event names are not mapped to an enum, it’s why this trick.
+                # Like CommandExecutionStateChangedEvent
+                try:
+                    raw_event_name = event.name.value
+                except AttributeError:
+                    raw_event_name = event.name
+
+                self.hass.bus.fire(
+                    f"Overkiz{raw_event_name}",
```
iMicknl, line 117: Perhaps Overkiz_{}?

```suggestion
                    f"Overkiz_{raw_event_name}",
```

iMicknl: Or maybe, make this more generic and add the specific event name to the payload. Or document the event names and how to catch them in specific scenario's.

tetienne: The name of the event is CameCase, it will be weird to prefix with `Overkiz_`

custom_components/tahoma/coordinator.py
```diff
@@ -98,6 +98,16 @@ async def _async_update_data(self) -> Dict[str, Device]:
         for event in events:
             _LOGGER.debug(event)
 
+            if event.failure_type_code:
+                self.hass.bus.fire(
+                    "overkiz.event",
```
tetienne, line 103: @iMicknl what do you think about?

```suggestion
                    "overkiz.failure",
```

Or we try to be generic and just publish everything?

iMicknl: Do you know how many events will be broadcasted if we publish everything? We don't want to fill the HA store with useless events / entries.

tetienne: If we publish everything, we will publish every events sent by Somfy. So most of the time there is nothing, except when you trigger an action. 
But for the moment, I don’t see a use case, except the failure one I gave as example. We can restrict the number of events today, and open later on demand.

iMicknl: Makes sense, let's just do it for this use-case and document it. (for example how to notify an user / create a persistent notification).

iMicknl (approved):

PR #578, opened 2021-09-23
Update awning detection

### LINKED ISSUE #577: PositionableScreen with reverse state

### The request

I have 3 SOMFY products for sun screening
2 * Awnings (PositionableHorizontalAwning)  that are implemented as expected 0% is closed and 100% is open
1* side screen (PositionableScreen) which is reversed in home assistant 0% open and 100% is closed

Connexoon

### Device model (optional)

_No response_

### Device type (optional)

_No response_

### Additional information

_No response_

tetienne:
I will strictly apply what Somfy Support told us: awning must use setDeployment. 
But some awnings have not the ui class set to awning (it will be too easy), like io:VerticalExteriorAwningIOComponent. So I now do a check more simple. 

Also our workaround for Valance is now useless, Somfy has updated their model with DeploymentState: https://www.tahomalink.com/enduser-mobile-web/enduserAPI/reference/controllable/io:AwningValanceIOComponent

Fix #577


custom_components/tahoma/cover.py
```diff
@@ -24,18 +24,16 @@ async def async_setup_entry(
     data = hass.data[DOMAIN][entry.entry_id]
     coordinator = data["coordinator"]
 
-    # Includes fix for #486, which is waiting on Somfy back-end deployment
-    # Remove when DeploymentState will be returned for AwningValance
     entities = [
         Awning(device.deviceurl, coordinator)
         for device in data["platforms"].get(COVER)
-        if device.ui_class == "Awning" and device.widget != "AwningValance"
+        if "Awning" in device.definition.qualified_name
```
iMicknl, line 30: Is it always in the qualified name? Previously we checked if it was part of the ui_class? Will this change be better?

iMicknl (approved):
LGTM. Let's wait with merging this one for some user feedback? Before we have 10 new issues again. 😀😀

custom_components/tahoma/cover.py
tetienne: From what I saw yes it’s always in the qualified name. I’ve tested myself, and ask [here](https://github.com/iMicknl/ha-tahoma/issues/577) and [here](https://github.com/iMicknl/ha-tahoma/issues/486)

tetienne: I’ve also checked all the issues about reversed state, the qualified name was always including Awning.

custom_components/tahoma/cover_devices/tahoma_cover.py
```diff
@@ -103,6 +103,9 @@ async def async_set_cover_tilt_position(self, **kwargs):
     def is_closed(self):
         """Return if the cover is closed."""
 
+        if self.current_cover_position is not None:
+            return self.current_cover_position == 0
```
tetienne, line 107: I will give priority to the position.

iMicknl: Makes sense, with the amount of errors in the Overkiz API, this would be more reliable.

custom_components/tahoma/cover.py
```diff
@@ -24,18 +24,16 @@ async def async_setup_entry(
...
     ]
 
     entities += [
         VerticalCover(device.deviceurl, coordinator)
         for device in data["platforms"].get(COVER)
-        if device.ui_class != "Awning" or device.widget == "AwningValance"
+        if "Awning" not in device.definition.qualified_name
```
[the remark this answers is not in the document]
tetienne: I would have to change also the class which is instanciate too. I can switch to a classic for loop instead of list comprenhension if you want.

iMicknl (approved):

PR #595, opened 2021-10-12
Handle also light for my position number

tetienne:
Follow up of #594


iMicknl:
Screen Shot 2021-10-12 at 12 28 29

Confirmed!

PR #608, opened 2021-10-16
Exclude Vertical Awning from new Awning fix

iMicknl:


iMicknl:
> Doing this you will break #577 where a user has a VerticalExteriorAwningIOComponent. What’s your device already?

I need to double check, but I recall I have a `io:VerticalExteriorAwningIOComponent - ExteriorScreen - PositionableScreen`.

PR #613, opened 2021-10-20
Change constants to centralized enum

iMicknl:
Inspired by https://github.com/home-assistant/core/blob/dev/homeassistant/components/tuya/const.py.

Shall we do this for the whole integration? Eventually we can move these enums to pyhoma as well, if we want.

<a href="https://gitpod.io/#https://github.com/iMicknl/ha-tahoma/pull/613"></a>


iMicknl:
Thanks @vlebourl for adding all enums :o. I ported them to https://github.com/iMicknl/python-tahoma-api/pull/228/files :)

PR #617, opened 2021-10-26
Improve support for Domestic Hot Water Production device

vlebourl:
* Fix away mode.
* Add boost mode.

<a href="https://gitpod.io/#https://github.com/iMicknl/ha-tahoma/pull/617"></a>

PR #637, opened 2021-10-29
Add more state sensors for DomesticHotWaterProduction

### LINKED ISSUE #568: Add more state sensors for DomesticHotWaterProduction (io:AtlanticDomesticHotWaterProductionV2_CE_FLAT_C2_IOComponent)

### The problem

The integration works well with cozytouch. I have a Thermor water heater which is well detected.
I can change mode etc

But an important data is missing. This data exists in the cozytouch custom component (https://github.com/Cyr-ius/hass-cozytouch). It is a boolean called energy_demand (io:OperatingModeCapabilitiesState energy_demand) which is set to true when water heater is heating.
I think it is beacause power is not retrieved from cozytouch data.
Some other data are missing like core:V40WaterVolumeEstimationState which is the hot water volume already consumed.

### What version of this integration (ha-tahoma) has the issue?

2.7.2

### What version of Home Assistant Core has the issue?

2021.9.6

CozyTouch

### Device model

DomesticHotWaterProduction

### Device type

E722804

### Additional information

_No response_

### Relevant log output

_No response_

iMicknl:
Fixes #568

<a href="https://gitpod.io/#https://github.com/iMicknl/ha-tahoma/pull/637"></a>

### DIFF
--- a/custom_components/tahoma/binary_sensor.py
+++ b/custom_components/tahoma/binary_sensor.py
@@ -78,7 +78,7 @@ BINARY_SENSOR_DESCRIPTIONS = [
         key="io:OperatingModeCapabilitiesState",
         name="Energy Demand Status",
         device_class=binary_sensor.DEVICE_CLASS_HEAT,
-        is_on=lambda state: state["energyDemandStatus"] == 0,
+        is_on=lambda state: state.get("energyDemandStatus") == 1,
     ),
 ]
 
--- a/custom_components/tahoma/binary_sensor.py
+++ b/custom_components/tahoma/binary_sensor.py
@@ -73,6 +73,13 @@ BINARY_SENSOR_DESCRIPTIONS = [
         device_class=binary_sensor.DEVICE_CLASS_VIBRATION,
         is_on=lambda state: state == STATE_DETECTED,
     ),
+    # DomesticHotWaterProduction/WaterHeatingSystem
+    OverkizBinarySensorDescription(
+        key="io:OperatingModeCapabilitiesState",
+        name="Energy Demand Status",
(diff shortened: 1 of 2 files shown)


custom_components/tahoma/binary_sensor.py
```diff
@@ -73,6 +73,13 @@
         device_class=binary_sensor.DEVICE_CLASS_VIBRATION,
         is_on=lambda state: state == STATE_DETECTED,
     ),
+    # DomesticHotWaterProduction/WaterHeatingSystem
+    OverkizBinarySensorDescription(
+        key="io:OperatingModeCapabilitiesState",
+        name="Energy Demand Status",
+        device_class=binary_sensor.DEVICE_CLASS_HEAT,
+        is_on=lambda state: state["energyDemandStatus"] == 0,
```
tetienne, line 81: Hopefully, for the moment we retrieve only one value from the json. If we need to retrieve a second value, we will have to rethink the description key system.

tetienne (approved):

custom_components/tahoma/binary_sensor.py
iMicknl: Same...

vlebourl:
Did you take a look at what #617 does? I'd like to merge it before we finish the enums as it fixes a bug in away mode and introduces boost mode (or the other way around).

iMicknl:
> Did you take a look at what #617 does? I'd like to merge it before we finish the enums as it fixes a bug in away mode and introduces boost mode (or the other way around).

You can merge that if is works :). This one will just add a few optional sensors, I don't think it has duplicate functionality?

PR #638, opened 2021-10-31
Move overkiz.event to the end of the `coordinator` logic

iMicknl:
@tetienne wouldn't it make sense to have this code at the end of our coordinator logic? This means that it won't interfere with our core functions at least.

<a href="https://gitpod.io/#https://github.com/iMicknl/ha-tahoma/pull/638"></a>

### DIFF
--- a/custom_components/tahoma/coordinator.py
+++ b/custom_components/tahoma/coordinator.py
@@ -102,16 +102,6 @@ class OverkizDataUpdateCoordinator(DataUpdateCoordinator):
         for event in events:
             _LOGGER.debug(event)
 
-            if event.failure_type_code:
-                self.hass.bus.fire(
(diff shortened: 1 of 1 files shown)


tetienne:
What do you mean by interfere?

iMicknl:
> What do you mean by interfere?

To make sure that issues like this https://github.com/iMicknl/ha-tahoma/issues/629, won't disable core functionality (like state etc.).

tetienne:
It should not happen anymore with my fix. But OK.

tetienne (approved):

iMicknl:
> It should not happen anymore with my fix. But OK.

I agree! However, it just made me think that we should try to have our most important logic before other logic.

PR #641, opened 2021-11-01
Fix number entity when state is None

iMicknl:
`core:Memorized1PositionState` isn't set, thus the state is not available. However, we create a number entity when they have the state definition available.  

This was the bug that @mstuij faced in https://github.com/iMicknl/ha-tahoma/issues/629#issuecomment-954811387. 

<a href="https://gitpod.io/#https://github.com/iMicknl/ha-tahoma/pull/641"></a>

### DIFF
--- a/custom_components/tahoma/number.py
+++ b/custom_components/tahoma/number.py
@@ -62,8 +62,8 @@ class OverkizNumber(OverkizDescriptiveEntity, NumberEntity):
     @property
     def value(self) -> float:
         """Return the current number."""
-        if self.device.states.get(self.entity_description.key):
-            return self.device.states.get(self.entity_description.key).value
+        if state := self.device.states.get(self.entity_description.key):
+            return state.value
 
         return None
 
--- a/custom_components/tahoma/number.py
+++ b/custom_components/tahoma/number.py
@@ -62,7 +62,10 @@ class OverkizNumber(OverkizDescriptiveEntity, NumberEntity):
     @property
     def value(self) -> float:
         """Return the current number."""
-        return self.device.states.get(self.entity_description.key).value
+        if self.device.states.get(self.entity_description.key):
+            return self.device.states.get(self.entity_description.key).value
+
+        return None
 
     async def async_set_value(self, value: float) -> None:
         """Update the My position value. Min: 0, max: 100."""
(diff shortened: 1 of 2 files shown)


custom_components/tahoma/select.py
```diff
@@ -59,7 +59,7 @@ def __init__(
     @property
     def current_option(self):
         """Return the selected entity option to represent the entity state."""
-        return self.device.states.get(CORE_OPEN_CLOSED_PEDESTRIAN_STATE).value
+        return self.device.states[CORE_OPEN_CLOSED_PEDESTRIAN_STATE].value
```
iMicknl, line 62: I also changed this, since it will throw an error anyways if it returns None. Thus it doesn't make sense to use `.get()` instead of calling it directly.

custom_components/tahoma/number.py
```diff
@@ -62,7 +62,10 @@ class OverkizNumber(OverkizDescriptiveEntity, NumberEntity):
     @property
     def value(self) -> float:
         """Return the current number."""
-        return self.device.states.get(self.entity_description.key).value
+        if self.device.states.get(self.entity_description.key):
+            return self.device.states.get(self.entity_description.key).value
```
tetienne, line 66: You can use the := operator

custom_components/tahoma/select.py
tetienne: I don't think it works it's not an array. It's why we use the get method.

tetienne (changes requested):

custom_components/tahoma/number.py
iMicknl: How would that look? :D

custom_components/tahoma/select.py
iMicknl: You could be right, I will revert this, since I cannot test it.

custom_components/tahoma/number.py
```diff
@@ -62,7 +62,10 @@ class OverkizNumber(OverkizDescriptiveEntity, NumberEntity):
     @property
     def value(self) -> float:
         """Return the current number."""
-        return self.device.states.get(self.entity_description.key).value
+        if state := self.device.states.get(self.entity_description.key):
+            return state.value
```
tetienne, line 66: @iMicknl done ;)

tetienne:
https://github.com/iMicknl/ha-tahoma/pull/603/files#r730304483 🙄

tetienne (approved):

PR #642, opened 2021-11-02
Improve select entity and add support for setting Simple Memorized Volume

iMicknl:
Gate logic hasn't been tested yet. Setting Simple Memorized Volume works.

<a href="https://gitpod.io/#https://github.com/iMicknl/ha-tahoma/pull/642"></a>


custom_components/tahoma/select.py
```diff
@@ -1,27 +1,36 @@
...
+            {
+                "closed": "close",
+                "open": "open",
+                "pedestrian": "setPedestrianPosition",
+            }[option]
+        ),
+    ),
+    OverkizSelectDescription(
+        key="io:MemorizedSimpleVolumeState",
+        name="Memorized Simple Volume",
```
tetienne, line 26: What is it?

custom_components/tahoma/select.py
```diff
@@ -33,40 +42,48 @@ async def async_setup_entry(
...
-    async_add_entities(entities)
+    key_supported_states = {
+        description.key: description for description in SELECT_DESCRIPTIONS
+    }
 
+    for device in coordinator.data.values():
+        for state in device.definition.states:
+            if description := key_supported_states.get(state.qualified_name):
+                print("added")
+                print(state.qualified_name)
```
tetienne, line 55: You forgot to remove this I guess

custom_components/tahoma/select.py
```diff
@@ -33,40 +42,48 @@ async def async_setup_entry(
...
-        super().__init__(device_url, coordinator)
-        self._attr_name = f"{super().name} Position"
+class OverkizSelect(OverkizDescriptiveEntity, SelectEntity):
+    """Representation of an Overkiz Number entity."""
 
     @property
     def current_option(self):
         """Return the selected entity option to represent the entity state."""
-        return self.device.states.get(CORE_OPEN_CLOSED_PEDESTRIAN_STATE).value
+        state = self.device.states.get(self.entity_description.key)
```
tetienne, line 73: Would be nice to use same code that the one in Number

custom_components/tahoma/select.py
iMicknl: Oops 🥺

custom_components/tahoma/select.py
iMicknl: Will have a look! I took this from sensor.oh

custom_components/tahoma/select.py
iMicknl: Like a My position; but for the audio volume

custom_components/tahoma/select.py
iMicknl: Fixed.

custom_components/tahoma/entity.py
```diff
@@ -144,6 +145,14 @@ class OverkizNumberDescription(NumberEntityDescription):
     state: Callable[[str], bool] = lambda state: state
 
 
+@dataclass
+class OverkizSelectDescription(SelectEntityDescription):
+    """Class to describe an Overkiz select entity."""
+
+    options: list[str] = None
+    select_option: Callable[[str], bool] = lambda option, execute_command: None
```
tetienne, line 153: I’m not sure about the execute_command parameter. In which case do you think it can be something else than the current execute_command?

We can add it later, if needed.

iMicknl: What do you mean? I need the execute_command parameter to call the function. See the code for the gate, I have to execute 3 different commands depending on the selected option (instead of just one command).

tetienne (approved):

PR #659, opened 2021-11-19
Remove unsupported device log

tetienne:
This log was only displayed when the debug level is enabled. It means most of the user does not see it by default, and think about when reading our README. And if they read it, it means they already now something is wrong with their setup. 

<a href="https://gitpod.io/#https://github.com/iMicknl/ha-tahoma/pull/659"></a>


custom_components/tahoma/const.py
```diff
@@ -99,7 +71,6 @@
     "Screen": COVER,
     "Shutter": COVER,
     "Siren": SWITCH,
-    "SirenStatus": None,  # widgetName, uiClass is Siren (switch)
```
iMicknl, line 102: We need to keep this, otherwise it will fallback to Switch (since uiClass siren is mapped to Switch).

custom_components/tahoma/const.py
```diff
@@ -37,34 +37,6 @@
     WATER_HEATER,
 ]
 
-IGNORED_OVERKIZ_DEVICES = [
-    "ProtocolGateway",
```
iMicknl, line 41: We should keep this, however we need to use this in a different way. In another PR I face issues where we create a device entry for `Pod` and `ProtocolGateway`, which we want to ignore.

However, the other entries could be removed indeed! :)

custom_components/tahoma/const.py
tetienne: It seems to be an hack. Can we fix it in someway?

iMicknl: I am not sure. It is an exception because Somfy uses the same ui class 🥺.

custom_components/tahoma/__init__.py
```diff
@@ -106,17 +105,15 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
...
+        )
         platform = OVERKIZ_DEVICE_TO_PLATFORM.get(
             device.widget
         ) or OVERKIZ_DEVICE_TO_PLATFORM.get(device.ui_class)
         if platform:
             platforms[platform].append(device)
-            log_device("Added device", device)
-        elif (
-            device.widget not in IGNORED_OVERKIZ_DEVICES
-            and device.ui_class not in IGNORED_OVERKIZ_DEVICES
```
Quentame, line 117: There is the same logic in platforms too.

Either :
- keep as it is
- remove all use of `IGNORED_OVERKIZ_DEVICES`

Quentame (changes requested):

iMicknl:
Fixed in core, lets fix this with a backport.

Issue #664, opened 2021-12-05
Missing boost mode and parameters for DomesticHotWaterProduction (io:AtlanticDomesticHotWaterProductionV2_SPLIT_IOComponent)


foXaCe:
### Did you read the instructions?

- [X] I have read the [README](https://github.com/iMicknl/ha-tahoma/blob/master/README.md), including the [advanced debugging](https://github.com/iMicknl/ha-tahoma/blob/master/README.md#advanced) section.

### The problem

missing the boost mode and some parameter appears null

![8377ce469c29fc4f657b4f48418bd5063d12a9b3](https://user-images.githubusercontent.com/290678/144758317-3a93b88e-522a-4d0b-8e4e-d9ba22474357.png)

### What version of this integration (ha-tahoma) has the issue?

2.9.4

### Which gateway / hub do you use?

Cozytouch

### Device model

DomesticHotWaterProduction par Atlantic Group Firmware: E722801

### Home Assistant log

```shell
[...]

PR #688, opened 2021-12-21
Add Boost Mode Duration and Away Mode Duration

iMicknl:


vlebourl:
Which DHWP device does this relate to? According to #617 commands differ depending on the device... I haven't found an elegant way to combine both listed in that PR yet...

iMicknl:
@vlebourl this is related to https://github.com/iMicknl/ha-tahoma/issues/664.

Perhaps it makes more sense to support BoostModeDuration only via a number entity, and not via the water heater? Since you have to set it for a number of days.

vlebourl:
Should we have a switch as well? app looks like this for boost mode for `io:AtlanticDomesticHotWaterProductionV2_CE_FLAT_C2_IOComponent`
image

foXaCe:
@iMicknl hello, is it possible to merge this function in master?
please

iMicknl:
> @iMicknl hello, is it possible to merge this function in master? please

We can look to add this feature to core, however this feature has not been tested very well. Which device do you have and do you use the away mode duration, or the boost mode duration?

foXaCe:
I miss the Boost function.
I can pass you my cozytouch password if you need to test.

iMicknl:
@foXaCe please contact me on Discord and we can have a look the coming weeks. I will see if we can add this feature to core.

iMicknl:
superseeded by https://github.com/home-assistant/core/pull/76690

PR #693, opened 2021-12-23
Clean code

tetienne:
Apply remarks from https://github.com/home-assistant/core/pull/62640

<a href="https://gitpod.io/#https://github.com/iMicknl/ha-tahoma/pull/693"></a>

### DIFF
--- a/custom_components/tahoma/sensor.py
+++ b/custom_components/tahoma/sensor.py
@@ -27,7 +27,6 @@ from homeassistant.helpers.entity_platform import AddEntitiesCallback
 from pyoverkiz.enums import OverkizAttribute, OverkizState, UIWidget
 
 from .const import DOMAIN, IGNORED_OVERKIZ_DEVICES
-from .coordinator import OverkizDataUpdateCoordinator
 from .entity import OverkizDescriptiveEntity, OverkizEntity, OverkizSensorDescription
 
 SENSOR_DESCRIPTIONS = [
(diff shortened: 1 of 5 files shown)


custom_components/tahoma/sensor.py
```diff
@@ -393,12 +392,9 @@ def state(self):
 class OverkizHomeKitSetupCodeSensor(OverkizEntity, SensorEntity):
     """Representation of an Overkiz HomeKit Setup Code."""
 
-    def __init__(self, device_url: str, coordinator: OverkizDataUpdateCoordinator):
-        """Initialize the device."""
-        super().__init__(device_url, coordinator)
-        self._attr_name = "HomeKit Setup Code"
-        self._attr_icon = "mdi:shield-home"
-        self._attr_entity_category = EntityCategory.DIAGNOSTIC
+    _attr_name = "HomeKit Setup Code"
```
iMicknl, line 395: This won't work, we need to set this in init, otherwise it will still be set to the label.

iMicknl: By the way, I broke HomeKit with ignored devices apparently. See my core PR, we need to move the if statement which sets up the entities.

tetienne: > This won't work, we need to set this in init, otherwise it will still be set to the label.

Ah yes you are right. I’m wondering if we cannot fix this.

iMicknl: I don't think so, since we set the label in init.

PR #696, opened 2021-12-29
Handle missing preset for AtlanticElectricalTowelDryer

tetienne:
See https://github.com/iMicknl/ha-tahoma/issues/682#issuecomment-1002781090

### DIFF
--- a/custom_components/tahoma/climate_devices/atlantic_electrical_towel_dryer.py
+++ b/custom_components/tahoma/climate_devices/atlantic_electrical_towel_dryer.py
@@ -88,9 +88,10 @@ class AtlanticElectricalTowelDryer(OverkizEntity, ClimateEntity):
     @property
     def preset_mode(self) -> Optional[str]:
         """Return the current preset mode, e.g., home, away, temp."""
-        return TAHOMA_TO_PRESET_MODE[
-            self.executor.select_state(IO_TOWEL_DRYER_TEMPORARY_STATE_STATE)
-        ]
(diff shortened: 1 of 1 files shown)


iMicknl:
@tetienne is this one still required?

tetienne:
Well it solves an issue. But without we don't break anything.

custom_components/tahoma/climate_devices/atlantic_electrical_towel_dryer.py
```diff
@@ -97,9 +97,10 @@ async def async_set_hvac_mode(self, hvac_mode: str) -> None:
     @property
     def preset_mode(self) -> Optional[str]:
         """Return the current preset mode, e.g., home, away, temp."""
-        return TAHOMA_TO_PRESET_MODE[
-            self.executor.select_state(IO_TOWEL_DRYER_TEMPORARY_STATE_STATE)
-        ]
+        state = self.executor.select_state(IO_TOWEL_DRYER_TEMPORARY_STATE_STATE)
```
iMicknl, line 100: Maybe we should just not add `SUPPORT_PRESET_MODE` when this state is not available?

tetienne: Can we do that in another PR? We will have to apply this logic to other devices too.

iMicknl: Why? Is that not only applicable to this device? I think the other climate devices always have the state present.

iMicknl:
Implemented in https://github.com/home-assistant/core/pull/73788/files

PR #700, opened 2022-01-05
Fix RTD indoor/outdoor siren

iMicknl:


iMicknl:
Fixed in core. Lets fix this with a backport.

PR #715, opened 2022-01-16
Bump PyOverkiz and move to Somfy OAuth 

iMicknl:
- [x] Merge PyOverkiz PR
- [x] Remove debug logging
- [x] Check reauth

<a href="https://gitpod.io/#https://github.com/iMicknl/ha-tahoma/pull/715"></a>

### DIFF
--- a/custom_components/tahoma/config_flow.py
+++ b/custom_components/tahoma/config_flow.py
@@ -87,7 +87,6 @@ class ConfigFlow(config_entries.ConfigFlow, domain=DOMAIN):
             except TooManyRequestsException:
                 errors["base"] = "too_many_requests"
             except BadCredentialsException as exception:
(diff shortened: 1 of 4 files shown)


custom_components/tahoma/config_flow.py
```diff
@@ -86,7 +86,8 @@ async def async_step_user(
                 return await self.async_validate_input(user_input)
             except TooManyRequestsException:
                 errors["base"] = "too_many_requests"
-            except BadCredentialsException:
+            except BadCredentialsException as exception:
+                _LOGGER.debug(type(exception))
```
iMicknl, line 90: @tetienne I used this temporarily so you can easily check if it uses new somfy auth. Just put in the wrong creds and it will say 'SomfyBadCredentialsException' in log, if new one is used.

Issue #697, opened 2021-12-30
Migrate ha-tahoma to Overkiz integration in core


iMicknl:
Let's keep track of the PR's and the work we need to do, to get everything to core.

**2022.2**
- [x] Base PR + Sensor
- [x] Lock PR + platforms
- [x] Button entity
- [x] Number entity
- [x] Light platform
- [x] Small Overkiz refactor
- [x] Add unique ID to config flow
- [x] Scene platform
- [x] Binary sensor entity
- [x] Address feedback: full mypy strict typing
- [x] Move to Overkiz 1.1.0 (remove casting in coordinator)
- [x] Select entity
- [x] Address feedback: improve coordinator with Registry
- [x] Don't close HA session in Config Flow
- [x] Switch platform
- [x] Update sensor platform
- [x] Add missing binary sensors
- [x] Fix where model can be str(none)
- [x] Add missing buttons 
- [x] Move to constants
- [x] Add diagnostics support
- [x] Address feedback (https://github.com/home-assistant/core/pull/64000#discussion_r786654864)
- [x] Move logger `_LOGGER = logging.getLogger(__name__)` to const.py and import it
- [x] Cover platform
- [x] Address feedback: better config flows tests (https://github.com/home-assistant/core/pull/62640#discussion_r776278803)
- [x] Implement reauth
- [x] Add zeroconf discovery
- [x] Implement Somfy OAuth

**2022.3**
- [x] Siren platform 
- [x] Support StatelessAlarmController and StatelessExteriorHeating via switch
- [x] Chain RTS cover commands
- [x] Add Low Speed functionality

**2022.4**
- [x] Alarm Control Panel platform (rewrite to EntityDescription)

**Climate implementations**
- [x] AtlanticElectricalHeater
- [x] AtlanticElectricalHeaterWithAdjustableTemperatureSetpoint https://github.com/home-assistant/core/pull/72790
- [x] AtlanticElectricalTowelDryer https://github.com/home-assistant/core/pull/73788
- [x] AtlanticHeatRecoveryVentilation https://github.com/home-assistant/core/pull/74015
- [x] AtlanticPassAPCHeatingAndCoolingZone
- [x] AtlanticPassAPCZoneControl
- [x] AtlanticPassAPCDHW
- [x] DimmerExteriorHeating (moved to number entity)
- [ ] EvoHomeController
- [ ] HeatingSetPoint
[...]

PR #717, opened 2022-01-18
Add support for ZeroConf discovery

egguy:
Hello,

I've added the support for discovering the overkiz box by ZeroConf.

The entry is: _kizbox._tcp.local.

There's a `properties` attribute which contain directly the gateway ref (aka `gateway_pin`).

<a href="https://gitpod.io/#https://github.com/iMicknl/ha-tahoma/pull/717"></a>

### DIFF
--- a/custom_components/tahoma/config_flow.py
+++ b/custom_components/tahoma/config_flow.py
@@ -6,7 +6,7 @@ from typing import Any
 
 from aiohttp import ClientError
 from homeassistant import config_entries
-from homeassistant.components import dhcp
+from homeassistant.components import dhcp, zeroconf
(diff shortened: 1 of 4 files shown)


iMicknl:
Thanks! Could you add a test as well? Since core has the requirement for 100% test coverage config flow, we require that for config flow as well.

If you want, you could also see how you can bring this to core yourselves as well. See https://github.com/iMicknl/ha-tahoma/issues/697. The config flow in core has some slight changes though.

iMicknl (approved):
Thanks! Would you like to do a PR to core as well, with this functionality? If you need help setting up core dev environment, I am happy to help.

PR #718, opened 2022-01-18
Prevent an error during the rename of a device

egguy:
When renaming a device on the app, it triggers a device reconfiguration, and the OverkizBinary sensor state is interrogated.

The problem seems that's the original device has not rediscovered yet. This trigger and exception until the referred device is discovered. 

This prevents the problem during the renaming of the device.

Perhaps a more long term solution would be to remove all the linked binary devices first ?  

```python
Error doing job: Task exception was never retrieved
  File "/usr/src/homeassistant/homeassistant/helpers/update_coordinator.py", line 134, in _handle_refresh_interval
    await self._async_refresh(log_failures=True, scheduled=True)
  File "/usr/src/homeassistant/homeassistant/helpers/update_coordinator.py", line 265, in _async_refresh
    update_callback()
  File "/usr/src/homeassistant/homeassistant/helpers/update_coordinator.py", line 325, in _handle_coordinator_update
    self.async_write_ha_state()
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 505, in async_write_ha_state
    self._async_write_ha_state()
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 538, in _async_write_ha_state
    state = self._stringify_state()
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 511, in _stringify_state
    if (state := self.state) is None:
  File "/usr/src/homeassistant/homeassistant/components/binary_sensor/__init__.py", line 205, in state
    return STATE_ON if self.is_on else STATE_OFF
  File "/config/custom_components/tahoma/binary_sensor.py", line 129, in is_on
    state = self.device.states.get(self.entity_description.key)
  File "/config/custom_components/tahoma/entity.py", line 50, in device
    return self.coordinator.data[self.device_url]
TypeError: 'NoneType' object is not subscriptable
```

<a href="https://gitpod.io/#https://github.com/iMicknl/ha-tahoma/pull/718"></a>


vlebourl:
Hi, thanks for the submission! I haven't tried any renaming but I think this might be a more general problem than just for method `is_on` of `binary_sensor`. 
We need to have wider look and fix this at a more general level.

tetienne:
@egguy Thx again for your contribution. Did you rename from an official Overkiz website/app (cozytouch, tahomalink, etc) or from HA?

egguy:
Yes it was after renaming on the official APP. This is the trigger:
```
2022-01-18 09:32:34 DEBUG (MainThread) [custom_components.tahoma.coordinator] Event(timestamp=1642494753205, name=<EventName.DEVICE_UPDATED: 'DeviceUpdatedEvent'>, setupoid=_CountingAttr(counter=375, _default=None, repr=<function obfuscate_id at 0x7fad1a0160>, eq=True, order=True, hash=None, init=True, on_setattr=None, metadata={}), owner_key=None, type=None, sub_type=None, time_to_next_state=None, failed_commands=None, failure_type_code=None, failure_type=None, condition_groupoid=None, place_oid='63b12451-77d5-47cb-9e66-987a3e3e0223', label='VR sejour', metadata=None, camera_id=None, deleted_raw_devices_count=None, protocol_type=None, gateway_id=None, exec_id=None, device_url=io://****-****-0891/13208893, device_states=[], old_state=None, new_state=None)
```
From `VR sejour` to `Volet séjour`

egguy:
I've renamed through the webportal (also happened with the app).
I've done so more test to reproduce, when renaming 2 entity quicky on the portal, it creates a mess:

```
Logger: homeassistant.config_entries
Source: helpers/entity_component.py:177
First occurred: 16:46:32 (13 occurrences)
Last logged: 16:46:32

Error unloading entry xxxxxx for scene
Error unloading entry xxxxxx for select
Error unloading entry xxxxxx for sensor
Error unloading entry xxxxxx for switch
Error unloading entry xxxxxx for water_heater
  File "/usr/src/homeassistant/homeassistant/config_entries.py", line 452, in async_unload
    result = await component.async_unload_entry(hass, self)  # type: ignore
  File "/usr/src/homeassistant/homeassistant/components/alarm_control_panel/__init__.py", line 118, in async_unload_entry
    return await component.async_unload_entry(entry)
  File "/usr/src/homeassistant/homeassistant/helpers/entity_component.py", line 177, in async_unload_entry
    raise ValueError("Config entry was never loaded!")
ValueError: Config entry was never loaded!
```

iMicknl:
I cannot reproduce this either and I think the problem is somewhere else. 

Renaming a device will still keep the same device_url, thus that shouldn't be the issue... (or I miss something).

iMicknl:
Let's close this one for now. If you can still reproduce with latest release, @egguy. Let's create an issue with a full debug log and see how we can tackle this :).
Thanks!

PR #722, opened 2022-01-21
Allow to chain RTS cover commands

tetienne:
Fix https://github.com/iMicknl/ha-tahoma/issues/720

<a href="https://gitpod.io/#https://github.com/iMicknl/ha-tahoma/pull/722"></a>

### DIFF
--- a/custom_components/tahoma/cover_devices/awning.py
+++ b/custom_components/tahoma/cover_devices/awning.py
@@ -54,8 +54,16 @@ class Awning(OverkizGenericCover):
 
     async def async_open_cover(self, **_):
         """Open the cover."""
-        await self.executor.async_execute_command(OverkizCommand.DEPLOY)
+        if self.device.device_url.startswith("rts"):
+            # Set the execution duration to 0 seconds
+            await self.executor.async_execute_command(OverkizCommand.DEPLOY, 0)
+        else:
+            await self.executor.async_execute_command(OverkizCommand.DEPLOY)
 
     async def async_close_cover(self, **_):
         """Close the cover."""
-        await self.executor.async_execute_command(OverkizCommand.UNDEPLOY)
+        if self.device.device_url.startswith("rts"):
+            # Set the execution duration to 0 seconds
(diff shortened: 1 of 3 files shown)


tetienne:
@iMicknl If this fix works, we will have to add the logic to the my button too.

tetienne:
User as confirmed the fix works.

custom_components/tahoma/cover_devices/awning.py
```diff
@@ -54,8 +54,16 @@ async def async_set_cover_position(self, **kwargs):
 
     async def async_open_cover(self, **_):
         """Open the cover."""
-        await self.executor.async_execute_command(OverkizCommand.DEPLOY)
+        if self.device.device_url.startswith("rts"):
+            # Set the execution duration to 0 seconds
+            await self.executor.async_execute_command(OverkizCommand.DEPLOY, 0)
```
iMicknl, line 59: Can we set this for io covers as well? Otherwise, could we create a new function `async_execute_cover_command`, that will handle this?

tetienne: It's only for rts

tetienne: This issue does not exist for io.

iMicknl: But I mean, can we always pass this parameter for cover?

tetienne: I don't think so as there is no parameters for io covers. Somfy returns an error in such case no?

tetienne: Just test it, we can add this parameter to io cover too, there is no error returned.

iMicknl: Great! Let's merge and test it, and than we can move it to core if no one complains!

tetienne:
Superseed by https://github.com/iMicknl/ha-tahoma/pull/780

PR #667, opened 2021-12-07
Fix number entity in 2022.1.0dev

iMicknl:
Fixes #665 

Broken by https://github.com/home-assistant/core/pull/61100.

(requires 2022.1.0dev)

<a href="https://gitpod.io/#https://github.com/iMicknl/ha-tahoma/pull/667"></a>

PR #726, opened 2022-01-26
Revert "Fix number entity in 2022.1.0dev"

tetienne:
Reverts iMicknl/ha-tahoma#667

This PR was merged too soon. Plus we will have to enfore the minimum required HA version. 

<a href="https://gitpod.io/#https://github.com/iMicknl/ha-tahoma/pull/726"></a>

### DIFF
--- a/custom_components/tahoma/number.py
+++ b/custom_components/tahoma/number.py
@@ -80,3 +80,13 @@ class OverkizNumber(OverkizDescriptiveEntity, NumberEntity):
         await self.executor.async_execute_command(
             self.entity_description.command, value
         )
+
+    @property
+    def min_value(self) -> float:
+        """Return the minimum value."""
+        return self.entity_description.min_value or self._attr_min_value
+
+    @property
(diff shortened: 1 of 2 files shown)


iMicknl:
Can we perhaps try to fix this based on the current HA version?

tetienne:
We added a new property, how can we make this backward compatible?

custom_components/tahoma/number.py
```diff
@@ -80,3 +80,13 @@ async def async_set_value(self, value: float) -> None:
...
+
+    @property
+    def min_value(self) -> float:
+        """Return the minimum value."""
+        return self.entity_description.min_value or self._attr_min_value
+
+    @property
+    def max_value(self) -> float:
+        """Return the maximum value."""
+        return self.entity_description.max_value or self._attr_max_value
```
iMicknl, line 92: I think we could check here for the HA version and otherwise just call super?

PR #734, opened 2022-01-26
Don't load integration if core integration is configured

iMicknl:
Screen Shot 2022-01-26 at 17 59 50

Not perfect since it will only check a boot / reload, however I think that would be good enough for now. Let's make sure that users don't configure both integrations.

<a href="https://gitpod.io/#https://github.com/iMicknl/ha-tahoma/pull/734"></a>

### DIFF
--- a/custom_components/tahoma/__init__.py
+++ b/custom_components/tahoma/__init__.py
@@ -50,7 +50,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
 
     if await _block_if_core_is_configured(hass, entry):
         raise ConfigEntryNotReady(
-            "You cannot use overkiz from core and the custom component at the same time."
+            "You cannot use Overkiz from core and custom component at the same time."
         )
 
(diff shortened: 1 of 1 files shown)


tetienne (dismissed):
I guess we will not backport this to the Core?

custom_components/tahoma/__init__.py
```diff
@@ -48,6 +48,11 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     password = entry.data[CONF_PASSWORD]
     server = SUPPORTED_SERVERS[entry.data[CONF_HUB]]
 
+    if await _block_if_core_is_configured(hass, entry):
+        raise ConfigEntryNotReady(
+            "You cannot use overkiz from core and the custom component at the same time."
```
tetienne, line 53:
```suggestion
            "You cannot use Overkiz from core and custom component at the same time."
```

iMicknl:
> I guess we will not backport this to the Core?

Nope! We just need it here :)

PR #741, opened 2022-01-31
Be more accurate on expected logs

tetienne:


iMicknl (approved):
Cool! Eventually I would love to change this to the new diagnostics feature.

tetienne:
I agree, we can do that once HA released 👌

PR #743, opened 2022-02-01
feat: Atlantic/Sauter Guelma DWH support

rguillard77:
Hi dear author,

I'm currently working on integrating a Sauter DHW which requires some additions to your project.
This DHW is derived from the Atlantic DomesticWaterHeater but with different states & operating modes.

I tried to implement a simple way to support any derived equipment models (Atlantic Group is known for OEM equipment) without relying to much on pythonic introspection and retaining as much as possible the current software architecture.

This PR relies on https://github.com/iMicknl/python-overkiz-api/pull/352 to be functional

Best,
R.

<a href="https://gitpod.io/#https://github.com/iMicknl/ha-tahoma/pull/743"></a>

### DIFF
--- a/custom_components/tahoma/water_heater.py
+++ b/custom_components/tahoma/water_heater.py
@@ -9,13 +9,21 @@ from .const import DOMAIN
 from .water_heater_devices.domestic_hot_water_production import (
     DomesticHotWaterProduction,
 )
+from .water_heater_devices.domestic_hot_water_production_sauter_guelma import (
+    DomesticHotWaterProductionSauterGUELMA,
+)
 from .water_heater_devices.hitachi_dhw import HitachiDHW
+from pyoverkiz.enums import OverkizState
 
 TYPE = {
     UIWidget.DOMESTIC_HOT_WATER_PRODUCTION: DomesticHotWaterProduction,
     UIWidget.HITACHI_DHW: HitachiDHW,
+    UIWidget.DOMESTIC_HOT_WATER_PRODUCTION + "SauterGUELMA": DomesticHotWaterProductionSauterGUELMA
 }
 
+def device_type(device):
+    device_full_name = device.widget + device.states.get(OverkizState.CORE_MANUFACTURER_NAME).value + device.states.get(OverkizState.CORE_NAME).value
+    return device_full_name if device_full_name in TYPE else device.widget
 
 async def async_setup_entry(
     hass: HomeAssistant,
@@ -31,7 +39,7 @@ async def async_setup_entry(
     ]
 
     entities = [
-        TYPE[device.widget](device.device_url, coordinator)
+        TYPE[device_type(device)](device.device_url, coordinator)
         for device in water_heater_devices
         if device.widget in TYPE
     ]
--- a/custom_components/tahoma/water_heater.py
+++ b/custom_components/tahoma/water_heater.py
@@ -3,7 +3,7 @@ from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import Platform
 from homeassistant.core import HomeAssistant
 from homeassistant.helpers.entity_platform import AddEntitiesCallback
-from pyoverkiz.enums import UIWidget
+from pyoverkiz.enums import OverkizState, UIWidget
 
 from .const import DOMAIN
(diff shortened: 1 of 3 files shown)


custom_components/tahoma/water_heater.py
```diff
@@ -3,19 +3,32 @@
...
+from .water_heater_devices.domestic_hot_water_production_sauter_guelma import (
+    DomesticHotWaterProductionSauterGUELMA,
+)
 from .water_heater_devices.hitachi_dhw import HitachiDHW
 
 TYPE = {
     UIWidget.DOMESTIC_HOT_WATER_PRODUCTION: DomesticHotWaterProduction,
     UIWidget.HITACHI_DHW: HitachiDHW,
+    UIWidget.DOMESTIC_HOT_WATER_PRODUCTION
+    + "SauterGUELMA": DomesticHotWaterProductionSauterGUELMA
```
iMicknl, line 21: I am not a huge fan of this + the device_type array. I think it will break other implementations that don't have a manufacturer name / core_name, and brings a lot of extra complexity.

Could we perhaps just look from the class itself which features we need to enable? I am not sure if these appliances are very different?

custom_components/tahoma/water_heater_devices/domestic_hot_water_production_sauter_guelma.py
```diff
@@ -0,0 +1,70 @@
...
+    """Representation of a DomesticHotWaterProductionSauterGUELMA Water Heater."""
+
+    _attr_operation_list = [*OPERATION_MODE_TO_OVERKIZ]
+
+    _attr_supported_features = SUPPORT_OPERATION_MODE | SUPPORT_AWAY_MODE
+    _attr_temperature_unit = TEMP_CELSIUS
+
+    def __init__(self, device_url, coordinator):
+        """Initialize DomesticHotWaterProductionSauterGUELMA"""
+        super().__init__(device_url, coordinator)
```
iMicknl, line 36:
```suggestion
```

iMicknl (commented):
Thanks for your contribution! Great to see more devices being added by an owner of a device, appreciated!

custom_components/tahoma/water_heater.py
rguillard77: I agree this is not ideal. I tried several implementations including having the DomesticHotWaterProduction class handle the device.
The major drawback is the addition of extra complexity while handling property 'is_away_mode_on' or callable to set the away mode of the device because my device (Sauter GUELMA) operating mode is different from Atlantic DHW.


Since the ui.widget state is identical to Genuine Atlantic DHW, using Manufacturer and Model state was my only way to discriminate the device to instantiate the right class.
I'm just beginning my journey with HA so I'll be glad to ear from you on how I can improve.
Best,

vlebourl:
regarding same UI.widget for different devices, we face the same problem here: https://github.com/iMicknl/ha-tahoma/pull/617
The solution is less than ideal, but it works with relatively low complexity. Maybe you could add your device to that implementation rather than having your own class? 
Branch is not yet merged, but it's actively worked on.

vlebourl:
> I'm just beginning my journey with HA so I'll be glad to ear from you on how I can improve.
Regarding the linter, you can install `pre-commit` and run it on your file to avoid the check to fail.
```
pip install pre-commit
pre-commit install
pre-commit run --all-files
```

rguillard77:
> regarding same UI.widget for different devices, we face the same problem here: #617 The solution is less than ideal, but it works with relatively low complexity. Maybe you could add your device to that implementation rather than having your own class? Branch is not yet merged, but it's actively worked on.

This resemble to my first implementation but I was not sure of the compliance with the project coding standards.
I'll do the changes to align the code with those guidelines. 
Do you want me to close this PR and re-submit another PR against the  feature/domestic_hot_water_production branch ?

rguillard77:
> > I'm just beginning my journey with HA so I'll be glad to ear from you on how I can improve.
> 
> Regarding the linter, you can install `pre-commit` and run it against your files to avoid the check to fail.
> 
> ```
> pip install pre-commit
> pre-commit install
> pre-commit run --all-files
> ```

Thanks, done !

PR #744, opened 2022-02-02
Backport Number platform from Core

tetienne:
See https://github.com/home-assistant/core/blob/dev/homeassistant/components/overkiz/number.py

<a href="https://gitpod.io/#https://github.com/iMicknl/ha-tahoma/pull/744"></a>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.8, 3.9]
+        python-version: ["3.9", "3.10"]
     steps:
       - uses: actions/checkout@v2
       - name: Set up Python
--- a/custom_components/tahoma/__init__.py
+++ b/custom_components/tahoma/__init__.py
@@ -1,11 +1,12 @@
 """The Overkiz (by Somfy) integration."""
+from __future__ import annotations
+
 import asyncio
 from collections import defaultdict
 from dataclasses import dataclass
 import logging
 
 from aiohttp import ClientError, ServerDisconnectedError
-from homeassistant.components.scene import DOMAIN as SCENE
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import CONF_PASSWORD, CONF_USERNAME, Platform
 from homeassistant.core import HomeAssistant, ServiceCall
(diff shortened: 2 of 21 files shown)


.github/workflows/test.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.8, 3.9]
+        python-version: [3.9, 3.10]
```
tetienne, line 11: Core only run the test on 3.9: https://github.com/home-assistant/core/blob/dev/.github/workflows/ci.yaml#L723

custom_components/tahoma/number.py
```diff
@@ -57,63 +75,53 @@
...
+    entities: list[OverkizNumber] = []
 
-    entities = []
-
-    key_supported_states = {
-        description.key: description for description in NUMBER_DESCRIPTIONS
-    }
-
-    for device in coordinator.data.values():
+    for device in data.coordinator.data.values():
```
tetienne, line 90: @iMicknl Coordinator is now a property of data?

custom_components/tahoma/cover.py
```diff
@@ -8,6 +8,8 @@
 from pyoverkiz.enums import UIClass
 import voluptuous as vol
 
+from custom_components.tahoma import HomeAssistantOverkizData
```
iMicknl, line 11: Should we just import from `.` here? (just like we do in core).

tetienne: I will mimic what we do on the Core side 👍

custom_components/tahoma/__init__.py
```diff
@@ -42,6 +44,15 @@
 SERVICE_EXECUTE_COMMAND = "execute_command"
 
 
+@dataclass
+class HomeAssistantOverkizData:
```
tetienne, line 48: Was a nice addition, avoid the magic string.

iMicknl: yeah I like it, took it from Tuya integration

requirements_dev.txt
```diff
@@ -1,2 +1,2 @@
 -r requirements.txt
-homeassistant==2021.12.0b6
+homeassistant==2022.2.0b6
```
iMicknl, line 2: Can you bump required version for HACS as well?

requirements_test.txt
```diff
@@ -1,5 +1,5 @@
 -r requirements_dev.txt
-pytest-homeassistant-custom-component==0.5.3
+pytest-homeassistant-custom-component==0.6.5
```
iMicknl, line 2:
```suggestion
pytest-homeassistant-custom-component==0.6.6
```

requirements_dev.txt
iMicknl, line 2:
```suggestion
homeassistant==2022.2.0
```

requirements_test.txt
tetienne: image
Meh

iMicknl (approved):
LGTM! (when tested locally)

tetienne:
@iMicknl I tested locally

PR #745, opened 2022-02-02
Add Target Closure Sensor

tetienne:
Fix https://github.com/iMicknl/ha-tahoma/issues/729

### DIFF
--- a/custom_components/tahoma/sensor.py
+++ b/custom_components/tahoma/sensor.py
@@ -328,6 +328,12 @@ SENSOR_DESCRIPTIONS = [
         key=OverkizState.IO_ELECTRIC_BOOSTER_OPERATING_TIME,
         name="Electric Booster Operating Time",
     ),
+    OverkizSensorDescription(
+        key=OverkizState.CORE_TARGET_CLOSURE,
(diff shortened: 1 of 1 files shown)


iMicknl:
Should we disable the sensor by default?

custom_components/tahoma/sensor.py
```diff
@@ -328,6 +328,12 @@
         key=OverkizState.IO_ELECTRIC_BOOSTER_OPERATING_TIME,
         name="Electric Booster Operating Time",
     ),
+    OverkizSensorDescription(
+        key=OverkizState.CORE_TARGET_CLOSURE,
+        name="Target Closure",
+        native_unit_of_measurement=PERCENTAGE,
+        entity_registry_enabled_default=False,
```
tetienne, line 335: @iMicknl It’s disabled by default to not confuse end users.

iMicknl: Sorry I wanted to do that indeed, but I missed that line apparently...

tetienne: No problem

iMicknl (approved):

PR #747, opened 2022-02-03
Add gateway_id to the config flow title.

vlebourl:


custom_components/tahoma/config_flow.py
```diff
@@ -124,36 +124,35 @@ async def async_step_reauth(
...
 
-    async def async_step_dhcp(self, discovery_info: dhcp.DhcpServiceInfo):
+    async def async_step_dhcp(self, discovery_info: dhcp.DhcpServiceInfo) -> FlowResult:
         """Handle DHCP discovery."""
         hostname = discovery_info.hostname
         gateway_id = hostname[8:22]
 
         _LOGGER.debug("DHCP discovery detected gateway %s", obfuscate_id(gateway_id))
-
-        if self._gateway_already_configured(gateway_id):
```
iMicknl, line 134: Could you remove this function as well? `_gateway_already_configured`

tests/test_config_flow.py
```diff
@@ -1,11 +1,14 @@
+"""Tests for Overkiz (by Somfy) config flow."""
 from __future__ import annotations
 
 from unittest.mock import Mock, patch
 
 from aiohttp import ClientError
-from homeassistant import config_entries, data_entry_flow, setup
+from homeassistant import config_entries, data_entry_flow
 from homeassistant.components import dhcp
+from homeassistant.components.overkiz.const import DOMAIN
```
iMicknl, line 9: Please change the tests for our custom integration ;-).

tests/test_config_flow.py
```diff
@@ -57,7 +60,7 @@ async def test_form(hass):
     with patch("pyoverkiz.client.OverkizClient.login", return_value=True), patch(
         "pyoverkiz.client.OverkizClient.get_gateways", return_value=None
     ), patch(
-        "custom_components.tahoma.async_setup_entry", return_value=True
+        "homeassistant.components.overkiz.async_setup_entry", return_value=True
```
iMicknl, line 63: Here as well (and all other ones)

custom_components/tahoma/config_flow.py
```diff
@@ -112,56 +124,54 @@ async def async_step_user(
...
-            _LOGGER.debug("Gateway %s is already configured", obfuscate_id(gateway_id))
-            return self.async_abort(reason="already_configured")
-
-        return await self.async_step_user()
+        _LOGGER.debug(
+            "DHCP discovery detected gateway %s on host %s",
+            obfuscate_id(gateway_id),
+            discovery_info.hostname,
```
iMicknl, line 135: Why this change? (and do we need to make this in core as well?)

iMicknl: But doesn't this print gateway-xxxxx? (so the raw information we use for gateway id?)

custom_components/tahoma/config_flow.py
```diff
@@ -112,56 +124,54 @@ async def async_step_user(
...
         self, discovery_info: zeroconf.ZeroconfServiceInfo
     ) -> FlowResult:
         """Handle ZeroConf discovery."""
 
         # abort if we already have exactly this bridge id/host
-        # reload the integration if the host got updated
         properties = discovery_info.properties
         gateway_id = properties["gateway_pin"]
 
         _LOGGER.debug(
```
vlebourl, line 148: Goes with this one as well.

custom_components/tahoma/config_flow.py
vlebourl: hmm possible! I was trying to print the ip address.

iMicknl: Hostname is not the ip address in this case. I think best would to remove both of these and just keep it the same as in core.

vlebourl: goes with 
```py
        _LOGGER.debug(
            "ZeroConf discovery detected gateway %s on host %s",
            obfuscate_id(gateway_id),
            discovery_info.host,
        )
```

iMicknl (approved):

PR #748, opened 2022-02-03
upgrade python version in gitpod.

vlebourl:


.gitpod.Dockerfile
```diff
@@ -2,4 +2,4 @@ FROM gitpod/workspace-full:latest
 
 USER gitpod
 
-RUN pip3 install black && echo "export PIP_USER=no" >> /home/gitpod/.bashrc
\ No newline at end of file
+RUN pyenv install 3.9.7 && pyenv global 3.9.7 && pip3 install black && echo "export PIP_USER=no" >> /home/gitpod/.bashrc
```
iMicknl, line 5: Weird that they just not load / provide a different container image... But apparently this is the way to go https://www.gitpod.io/docs/languages/python#python-versions.

iMicknl (approved):

PR #750, opened 2022-02-04
Fix core detection when overkiz is ignored

vlebourl:


custom_components/tahoma/__init__.py
```diff
@@ -222,11 +222,11 @@ def log_device(message: str, device: Device) -> None:
 async def _block_if_core_is_configured(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     overkiz_config_entries = hass.config_entries.async_entries("overkiz")
 
-    for overkiz_entry in overkiz_config_entries:
-        if (
-            entry.data[CONF_USERNAME] == overkiz_entry.data[CONF_USERNAME]
+    return any(
+        (
+            overkiz_entry.source != "ignore"
```
tetienne, line 227: Most of the time, there is always a constant, like [here](https://github.com/home-assistant/core/blob/494ef2f9b2e5feeb10dce94a02ee90a9a7430b6a/homeassistant/config_entries.py#L54)

tetienne:
Good catch

custom_components/tahoma/__init__.py
vlebourl: can be several things depending on how the user added the integration. For us, I guess one of those:
```py
SOURCE_USER = "user"
SOURCE_ZEROCONF = "zeroconf"
SOURCE_DHCP = "dhcp"
```

vlebourl, line 227:
```suggestion
            overkiz_entry.source in [SOURCE_USER, SOURCE_ZEROCONF, SOURCE_DHCP]
```

vlebourl: Is that ok with you?

custom_components/tahoma/__init__.py
```diff
@@ -222,11 +227,11 @@ def log_device(message: str, device: Device) -> None:
 async def _block_if_core_is_configured(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     overkiz_config_entries = hass.config_entries.async_entries("overkiz")
 
-    for overkiz_entry in overkiz_config_entries:
-        if (
-            entry.data[CONF_USERNAME] == overkiz_entry.data[CONF_USERNAME]
+    return any(
+        (
+            overkiz_entry.source in [SOURCE_USER, SOURCE_ZEROCONF, SOURCE_DHCP]
```
iMicknl, line 232:
```suggestion
            overkiz_entry.source in [SOURCE_USER, SOURCE_ZEROCONF, SOURCE_DHCP]
```

When you set up an entry (even when clicked via zeroconf/dhcp), doesn't it always use source user? Wouldn't this current change detect if a hub has been discovered but not set up as well?

vlebourl: From my testing, hubs that haven't been set up ar not in `hass.config_entries.async_entries("overkiz")`. And the source is kept as `zeroconf` (or `dhcp` I guess but untested) if you set up a hub that was auto discovered. I think to get `user`, you need to actively select "Add Integration" then "No, set up another instance of Overkiz (by Somfy)".

vlebourl: image

iMicknl: Thanks! It is always a bit vague how that behaves, thanks for double checking.

iMicknl (approved):

custom_components/tahoma/__init__.py
tetienne: @vlebourl Why did you change your mind?

PR #763, opened 2022-02-11
Reintroduce low speed feature

tetienne:
This PR reintroduces the low speed feature supported by some covers.

As you can see, it’s not anymore the service nor the switch. IMO, bot would be rejected by the Core team:
 * Virtual entity are not allowed
 * Our service was based on a new feature flag (1024)

<a href="https://gitpod.io/#https://github.com/iMicknl/ha-tahoma/pull/763"></a>

### DIFF
--- a/custom_components/tahoma/cover.py
+++ b/custom_components/tahoma/cover.py
@@ -1,5 +1,5 @@
 """Support for Overkiz covers - shutters etc."""
-from pyoverkiz.enums import UIClass
+from pyoverkiz.enums import OverkizCommand, UIClass
 
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import Platform
@@ -29,6 +29,13 @@ async def async_setup_entry(
         VerticalCover(device.device_url, data.coordinator)
         for device in data.platforms[Platform.COVER]
         if device.ui_class != UIClass.AWNING
+    ]
+
+    entities += [
+        VerticalCover(device.device_url, data.coordinator, low_speed=True)
+        for device in data.platforms[Platform.COVER]
+        if device.ui_class != UIClass.AWNING
+        and OverkizCommand.SET_CLOSURE_AND_LINEAR_SPEED in device.definition.commands
     ]
 
     async_add_entities(entities)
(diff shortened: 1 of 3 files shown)


vlebourl:
So low speed cannot be deactivated now?

custom_components/tahoma/cover.py
```diff
@@ -31,4 +31,11 @@ async def async_setup_entry(
         if device.ui_class != UIClass.AWNING
     ]
 
+    entities += [
+        VerticalCover(device.device_url, data.coordinator, low_speed=True)
+        for device in data.platforms[Platform.COVER]
+        if device.ui_class != UIClass.AWNING
+        and OverkizCommand.SET_CLOSURE_AND_LINEAR_SPEED in device.definition.commands
+    ]
+
```
vlebourl, line 40: Does this overrides the entities previously added for the ones having the low speed command?

tetienne: Nope, an other one is added. So the user will have two cover entities for the same device.

tetienne:
> So low speed cannot be deactivated now?

Yes, but It was already the case with the switch or the service before the backport.

Mariusthvdb:
why would a switch be rejected by the core team? as a matter of fact, a switch was introduced yesterday in dev for the Opentherm comfort mode on Plugwise, which seems an identical situation?

tetienne:
Are you talking about this https://github.com/home-assistant/core/pull/66261/files?

The switch here is mapped to something returned by the API. It's mean that to know the state of the switch you can call the API.

For the Low speed feature, there is no such thing. Somfy does expose not a state mapped to it.

It's why I talk about virtual switch. Behind the scene I stored its state, and never retrieve it from Somfy. And this is forbidden by the Core team.

Mariusthvdb:
Yes that the one.

I see, that is indeed very different. Didn't realize Somfy didn't expose it in their api. (Might be a good FR for Somfy... )

Does this mean we have to create the switch let's say as an inout_boolean, and configure a template cover based on that?

tetienne:
That's indeed a good idea

PR #765, opened 2022-02-20
Backport `binary_sensor` entity

iMicknl:


custom_components/tahoma/binary_sensor.py
```diff
@@ -1,147 +1,161 @@
...
-        name="Boost Mode",
-        icon="hass:water-boiler-alert",
-        is_on=lambda state: state == OverkizCommandParam.ON,
-    ),
-    OverkizBinarySensorDescription(
-        key=OverkizState.IO_DHW_ABSENCE_MODE,
-        name="Away Mode",
-        icon="hass:water-boiler-off",
-        is_on=lambda state: state == OverkizCommandParam.ON,
-    ),
```
tetienne, line 95: We should keep these ones

iMicknl: Fixed

PR #768, opened 2022-02-20
Backport `sensor` entity

iMicknl:


custom_components/tahoma/sensor.py
```diff
@@ -24,38 +28,49 @@
...
+
+SENSOR_DESCRIPTIONS: list[OverkizSensorDescription] = [
     OverkizSensorDescription(
         key=OverkizState.CORE_BATTERY_LEVEL,
         name="Battery Level",
         native_unit_of_measurement=PERCENTAGE,
         device_class=SensorDeviceClass.BATTERY,
         state_class=SensorStateClass.MEASUREMENT,
         entity_category=EntityCategory.DIAGNOSTIC,
+        native_value=lambda value: int(float(str(value).strip("%"))),
```
tetienne, line 56: Really?

custom_components/tahoma/sensor.py
```diff
@@ -78,13 +93,6 @@
         entity_registry_enabled_default=False,
         state_class=SensorStateClass.MEASUREMENT,
     ),
-    OverkizSensorDescription(
```
tetienne, line 81: We should keep it, I guess.

custom_components/tahoma/sensor.py
```diff
@@ -111,24 +119,18 @@
...
         state_class=SensorStateClass.MEASUREMENT,
+        native_unit_of_measurement=TEMP_CELSIUS,
     ),
     OverkizSensorDescription(
         key=OverkizState.IO_MIDDLE_WATER_TEMPERATURE,
         name="Middle Water Temperature",
         device_class=SensorDeviceClass.TEMPERATURE,
         state_class=SensorStateClass.MEASUREMENT,
-    ),
-    OverkizSensorDescription(
```
tetienne, line 121: This one also.

custom_components/tahoma/sensor.py
iMicknl: This is part of core haha, I know it could be nicer. However it really sucks that this is not standardized and some contain `100%` as string.

PR #776, opened 2022-02-20
Backport entity

iMicknl:
(requires binary_sensor and sensor to be merged first)

### DIFF
--- a/custom_components/tahoma/entity.py
+++ b/custom_components/tahoma/entity.py
@@ -1,15 +1,13 @@
 """Parent class for every Overkiz device."""
 from __future__ import annotations
 
-from dataclasses import dataclass
-from typing import Any, Callable
+from enum import unique
+from typing import cast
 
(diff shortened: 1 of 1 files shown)


custom_components/tahoma/entity.py
```diff
@@ -71,56 +65,24 @@ def device_info(self) -> DeviceInfo:
...
             ),
+            hw_version=self.device.controllable_name,
             suggested_area=self.coordinator.areas[self.device.place_oid],
-            via_device=self.executor.get_gateway_id(),
+            via_device=(DOMAIN, self.executor.get_gateway_id()),
             configuration_url=self.coordinator.client.server.configuration_url,
         )
 
-    @property
-    def extra_state_attributes(self) -> dict[str, Any]:
```
tetienne, line 91: We will have to add this as breaking change.

iMicknl: yeah, there are more. I will add some text to the changelog! :)

PR #777, opened 2022-02-20
Refactor Alarm Control Panel

iMicknl:


custom_components/tahoma/alarm_entities/tsk_alarm_controller.py
```diff
@@ -0,0 +1,77 @@
...
+        if self.executor.select_state(
+            OverkizState.INTERNAL_CURRENT_ALARM_MODE
+        ) != self.executor.select_state(OverkizState.INTERNAL_TARGET_ALARM_MODE):
+            return STATE_ALARM_PENDING
+
+        MAP_INTERNAL_STATUS_STATE = {
+            OverkizCommandParam.OFF: STATE_ALARM_DISARMED,
+            "zone1": STATE_ALARM_ARMED_HOME,
+            "zone2": STATE_ALARM_ARMED_NIGHT,
+            OverkizCommandParam.TOTAL: STATE_ALARM_ARMED_AWAY,
```
iMicknl, line 50: I will need to add `zone1` and `zone2` to pyoverkiz. This is not according to Somfy spec. It should be `partial1` and `partial2`, but in my and other users tests it seems that they use zone instead of partial.

tetienne (approved):

PR #780, opened 2022-02-22
Add default delay of 0 to RTS commands

iMicknl:
Fixes https://github.com/iMicknl/ha-tahoma/issues/720

### DIFF
--- a/custom_components/tahoma/executor.py
+++ b/custom_components/tahoma/executor.py
@@ -15,10 +15,10 @@ from .coordinator import OverkizDataUpdateCoordinator
 # the delay to another value
 COMMANDS_WITHOUT_DELAY = [
     OverkizCommand.IDENTIFY,
-    "test",
+    OverkizCommand.TEST,
     OverkizCommand.ON,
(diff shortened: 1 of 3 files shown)


custom_components/tahoma/executor.py
```diff
@@ -4,13 +4,23 @@
...
 from pyoverkiz.types import StateType as OverkizStateType
 
 from .const import LOGGER
 from .coordinator import OverkizDataUpdateCoordinator
 
+# Commands that don't support setting
+# the delay to another value
+COMMANDS_WITHOUT_DELAY = [
+    OverkizCommand.IDENTIFY,
+    "test",
```
tetienne, line 18: I guess you will use the new pymfy version to replace this string?

iMicknl: Yes! :D

iMicknl: Done

tetienne (approved):

vlebourl (approved):

PR #790, opened 2022-03-08
Emulate open/close_cover_tilt via set_cover_tilt_position

greg2001:
Makes the UI nicer.

Before:
2022-03-08 13_19_06-Window

After:
2022-03-08 13_22_35-Window


tetienne:
Did you check on the official HA Discord if this behavior can be integrated into the cover entity directly? If we follow the logic the same behavior must be applied for open and close cover.

Can you also share your diagnostic file, please?

iMicknl:
> Just let me know, what kind of a diagnostic do you need and how do I get it, I'll gladly send it to you.

You can download this via the 3 dots on the integration page. Just select 'download diagnostics'.

PR #791, opened 2022-03-09
Add core integration usage to README

iMicknl:
I removed the HACs button since i don't think it adds anything. Let me know if you think the same.


tetienne (approved):
Well it's OK for me

PR #797, opened 2022-04-07
Select correct coordinator for execute_command service

vlebourl:
In case several hubs are registered, the service was called on the latest coordinator registered, which could thus be one not corresponding to the entity passed as argument. This fix allows to retrieve the coordinator attached to the given entity.

can relate to https://github.com/iMicknl/ha-tahoma/issues/167

### DIFF
--- a/custom_components/tahoma/__init__.py
+++ b/custom_components/tahoma/__init__.py
@@ -158,10 +158,16 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
         entity_registry = await hass.helpers.entity_registry.async_get_registry()
 
         for entity_id in call.data.get("entity_id"):
-            entity = entity_registry.entities.get(entity_id)
+            registry_entry = entity_registry.entities.get(entity_id)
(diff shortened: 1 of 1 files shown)


vlebourl:
can we merge ?

custom_components/tahoma/__init__.py
```diff
@@ -158,10 +158,15 @@ async def handle_execute_command(call: ServiceCall):
         entity_registry = await hass.helpers.entity_registry.async_get_registry()
 
         for entity_id in call.data.get("entity_id"):
-            entity = entity_registry.entities.get(entity_id)
+            registry_entry = entity_registry.entities.get(entity_id)
+            entity = [
+                e
+                for e in hass.data["entity_components"][registry_entry.domain].entities
+                if e.unique_id == registry_entry.unique_id
+            ][0]
```
iMicknl, line 166: You could perhaps also use `next()` here, instead of `[0]`.

iMicknl (approved):
Nice fix!

PR #825, opened 2022-11-03
Bugfixes for 2022.11

iMicknl:


custom_components/tahoma/select.py
```diff
@@ -21,7 +21,6 @@
 class OverkizSelectDescriptionMixin:
     """Define an entity description mixin for select entities."""
 
-    options: list[str | OverkizCommandParam]
```
tetienne, line 24: How does it work now?

tetienne (dismissed):

custom_components/tahoma/select.py
iMicknl: They added options to the native core description.

vlebourl (approved):

